PR #5, opened 2013-08-31
[JENKINS-19409] - Added option, which makes plugin to convert all SIDs to the lower-case

oleg-nenashev:
Resolves: https://issues.jenkins-ci.org/browse/JENKINS-19409

Signed-off-by: Oleg Nenashev nenashev@synopsys.com

PR #10, opened 2015-04-14
Add case sensitive help

fbelzunc:
Users still try to add groups in Jenkins which are not exactly written in the same way they are using in their Access Control, e.g. LDAP, AD, ... 

Hopefully, some of them will read the help button and will figure out that it is indeed case sensitive and it needs to be written in the same way.

@reviewbybees

### DIFF
--- a/src/main/webapp/help/help-user-group-add.html
+++ b/src/main/webapp/help/help-user-group-add.html
@@ -1,0 +1,7 @@
+<div>
+    <p>
+        Add users and groups. By default user/group names are case sensitive so it
+        needs to match with your user/group definition under Configure Global
+        Security->Access Control->Security Realm.
+    </p>
+</div>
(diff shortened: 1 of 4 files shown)


src/main/webapp/help/help-user-group-add.html
```diff
@@ -0,0 +1,31 @@
+<!--
+  - The MIT License
+  -
+  - Copyright (c) 2010, Manufacture Française des Pneumatiques Michelin, Thomas Maurel
```
KostyaSha: Are you sure?

oleg-nenashev: Wrong copyright

fbelzunc: Ups. Right. Just copy/paste sorry. I am modifying it.

oleg-nenashev:
The text is good for me. BTW, we still need a case-insensitive mode (e.g. draft implementation from https://github.com/jenkinsci/role-strategy-plugin/pull/5)

fbelzunc:
@oleg-nenashev @KostyaSha I have just removed the copyright.

oleg-nenashev:
Thanks for your contribution!
Please merge changes into a single commit. I'll merge the PR after it.

fbelzunc:
@oleg-nenashev done!

PR #11, opened 2015-04-20
show dialog box to confirm user input starting or ending with whitespace characters to avoid failing matches

holgerfriedrich:
We often had rules which did not match because of unwanted tailing whitespace after user names and patterns. Once noticed, one can not edit this in the GUI but needs to recreate the correct item or resolve it in config.xml.

This pull request removes whitespace during the form processing:

manage roles: trim pattern to avoid pattern mismatch due to tailing whitespace
assign roles: trim user/group name to avoid mismatch due to tailing whitespace

### DIFF
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
@@ -427,7 +427,7 @@ public class RoleBasedAuthorizationStrategy extends AuthorizationStrategy {
               if(e.getValue()) {
                 Role role = roleMap.getRole(e.getKey());
                 if(role != null && sid != null && !sid.equals("")) {
-                  roleMap.assignRole(role, sid);
+                  roleMap.assignRole(role, sid.trim());
                 }
               }
             }
@@ -454,7 +454,7 @@ public class RoleBasedAuthorizationStrategy extends AuthorizationStrategy {
 
         JSONObject globalRoles = formData.getJSONObject(GLOBAL);
         for(Map.Entry<String,JSONObject> r : (Set<Map.Entry<String,JSONObject>>)globalRoles.getJSONObject("data").entrySet()) {
-          String roleName = r.getKey();
+          String roleName = r.getKey().trim();
           Set<Permission> permissions = new HashSet<Permission>();
           for(Map.Entry<String,Boolean> e : (Set<Map.Entry<String,Boolean>>)r.getValue().entrySet()) {
               if(e.getValue()) {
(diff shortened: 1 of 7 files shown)


KostyaSha:
:-1: space can be valid character in AD/LDAP

KostyaSha:
> Once noticed, one can not edit this in the GUI but needs to recreate the correct item or resolve it in config.xml.

This imho need to be extended to allow renaming

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -427,7 +427,7 @@ public void doAssignSubmit(StaplerRequest req, StaplerResponse rsp) throws Unsup
               if(e.getValue()) {
                 Role role = roleMap.getRole(e.getKey());
                 if(role != null && sid != null && !sid.equals("")) {
-                  roleMap.assignRole(role, sid);
+                  roleMap.assignRole(role, sid.trim());
```
KostyaSha: Logic looks weird, if statement checks that sid is not empty, so space is valid, and then after trim spaces can be degenerated to `""` that where not valid before.

holgerfriedrich:
spaces are ok in usernames, but does it make sense to allow spaces before or after a user name?

holgerfriedrich:
you are right, should be trimmed before this check...

holgerfriedrich:
can we at least remove the tailing spaces for pattern?

KostyaSha:
Don't matter if value can be valid. The same situation is for job names, they are trimmed but in rename you can change, this leads to hidden magic. You entered value, it accepted but saved with other entry. 
I will prefer to see some red highlight for starting/ending spaces in UI. 
This is my imho, you need wait maintainer.

holgerfriedrich:
Highlighting would also be fine for me....

holgerfriedrich:
Rolled back changes in the java file as suggested.
Added a confirmation dialog shown after the user presses the add button in case the input starts or ends with a whitespace character.
This way, existing configurations are not changed, no hidden magic is done, and the user can enter it if it is really intended.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
oleg-nenashev: Yep, needs a refactoring

oleg-nenashev:
@holgerfriedrich @KostyaSha 
Hi Holger and Kostya, sorry for missing this PR and for joining too late. I think that it would be better to just prohibit the creation of new roles with trailing spaces and to probably print administrative warnings for existing ones. I don't think that empty roles and trailing spaces are the realistic use-case even if LDAP/AD support it. This imposes a risk for the internal code. BTW the approach with confirmations is acceptable as well, because they also act as disclaimers.

I hope to spend some time on the PRs integration on early December. @KostyaSha can go forward and merge the PR if he feels that it's ready

darxriggs:
ping

PR #13, opened 2015-10-30
Treat user authorities as roles

bkmeneguello:
This is useful in combination with other plugins, like LDAP Authentication to treat groups as roles.
Also I've added two tests. One covering the basic working flow and another to check if authorities are being treated as roles.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -77,6 +77,24 @@
             <version>1.14</version>
             <type>jar</type>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>1.8.5</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-module-junit4</artifactId>
+            <version>1.4.10</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-api-mockito</artifactId>
+            <version>1.4.10</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>  
   
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
@@ -40,8 +40,15 @@ import java.util.SortedMap;
 import java.util.SortedSet;
 import java.util.TreeMap;
 import java.util.TreeSet;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import java.util.regex.Matcher;
+import jenkins.model.Jenkins;
+import org.acegisecurity.BadCredentialsException;
+import org.acegisecurity.GrantedAuthority;
 import org.acegisecurity.acls.sid.Sid;
+import org.acegisecurity.userdetails.UserDetails;
+import org.springframework.dao.DataAccessException;
 
 /**
  * Class holding a map for each kind of {@link AccessControlled} object, associating
@@ -52,6 +59,8 @@ public class RoleMap {
 
   /** Map associating each {@link Role} with the concerned {@link User}s/groups. */
   private final SortedMap <Role,Set<String>> grantedRoles;
+
+  private static final Logger LOGGER = Logger.getLogger(RoleMap.class.getName());
 
   RoleMap() {
     this.grantedRoles = new TreeMap<Role, Set<String>>();
@@ -66,9 +75,9 @@ public class RoleMap {
    * @return True if the sid's granted permission
    */
   private boolean hasPermission(String sid, Permission p, RoleType roleType, AccessControlled controlledItem) {
-    for(Role role : getRolesHavingPermission(p)) {     
+    for(Role role : getRolesHavingPermission(p)) {
         
-        if(this.grantedRoles.get(role).contains(sid)) {            
+        if(this.grantedRoles.get(role).contains(sid)) {
             // Handle roles macro
             if (Macro.isMacro(role)) {
                 Macro macro = RoleMacroExtension.getMacro(role.getName());
@@ -76,14 +85,27 @@ public class RoleMap {
                     RoleMacroExtension macroExtension = RoleMacroExtension.getMacroExtension(macro.getName());
                     if (macroExtension.IsApplicable(roleType) && macroExtension.hasPermission(sid, p, roleType, controlledItem, macro)) {
                         return true;
-                    }          
+                    }
                 }
             } // Default handling
             else {
                 return true;
             }
+        } else {
+            try {
+                UserDetails userDetails = Jenkins.getInstance().getSecurityRealm().loadUserByUsername(sid);
+                for (GrantedAuthority grantedAuthority : userDetails.getAuthorities()) {
+                    if (grantedAuthority.getAuthority().equals(role.getName())) {
+                        return true;
+                    }
+                }
+            } catch (BadCredentialsException e) {
+                LOGGER.log(Level.FINE, "Bad credentials", e);
+            } catch (DataAccessException e) {
+                LOGGER.log(Level.FINE, "failed to access the data", e);
+            }
         }
-                    
+
         // TODO: Handle users macro
     }
     return false;
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
@@ -24,6 +24,9 @@
 
 package com.michelin.cio.hudson.plugins.rolestrategy;
 
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.collect.MapMaker;
 import com.synopsys.arc.jenkins.plugins.rolestrategy.Macro;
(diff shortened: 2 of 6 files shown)


src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -76,14 +79,23 @@ private boolean hasPermission(String sid, Permission p, RoleType roleType, Acces
...
             }
+        } else {
+            try {
+                UserDetails userDetails = Jenkins.getInstance().getSecurityRealm().loadUserByUsername(sid);
+                for (GrantedAuthority grantedAuthority : userDetails.getAuthorities()) {
+                    if (grantedAuthority.getAuthority().equals(role.getName())) {
+                        return true;
+                    }
+                }
+            } catch (Exception e) {}
```
oleg-nenashev: There should be logging at least

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -76,14 +79,23 @@ private boolean hasPermission(String sid, Permission p, RoleType roleType, Acces
...
-                    }          
+                    }
                 }
             } // Default handling
             else {
                 return true;
             }
+        } else {
+            try {
+                UserDetails userDetails = Jenkins.getInstance().getSecurityRealm().loadUserByUsername(sid);
```
oleg-nenashev: Have you evaluated the performance impact? If I understand the code correctly. it's going to cause a huge performance degradation. Caching is required IMO.

src/main/resources/org/jenkinsci/plugins/rolestrategy/RoleAssignmentTest/testRoleAssignment/config.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<hudson>
```
oleg-nenashev: Test resources should be located in src/test/resources, not in main

oleg-nenashev:
@bkmeneguello 
Hi Bruno,

I'm aware about the performance impact of the change. The change is going to embedded user loading from security realms and sequential role checks for every SID missing in the group. Since it's a common case for the fine-grain security, the overall calculation time may seriously degrade. 

Have you evaluated it?

pom.xml
```diff
@@ -77,6 +77,24 @@
             <version>1.14</version>
             <type>jar</type>
         </dependency>
+		<dependency>
```
KostyaSha: wrong intend

oleg-nenashev: It would be better to solve architectural issues before rat-holing into this topic...

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
bkmeneguello: Sure. I just preserved the original behaviour of doing nothing when the role verification fails. But I admit that using Exception is too wide. I'll narrow it. You sure logging could be a good idea?

src/main/resources/org/jenkinsci/plugins/rolestrategy/RoleAssignmentTest/testRoleAssignment/config.xml
bkmeneguello: Ops.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
bkmeneguello: I really didn't. I'm using this change in my local environment for evaluation. We have 40 users actively working with Jenkins and, until now, didn't noticed any performance penalty. Our security realm is LDAPSecurityRealm without auth cache. It's performing very well.
But I agree that other security realms may perform poor. Do you suggest some attention points that can be bottlenecks? Thanks.

pom.xml
bkmeneguello: Fixed. Thanks!

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
bkmeneguello: Do you think this feature should be an opt-in?

bkmeneguello:
@oleg-nenashev So do you had reviewed my last comments?
I'm using this changed plugin a while in my company and didn't noticed any performance issues. Whe have near 40 users and 60 JOBs with project and general roles.
Do you have any attention point that I should consider in particular?
Thanks.

bkmeneguello:
Ok, today I got what you mean. My server become unusable because of a new LDAP ACL that caused slowness.
I'll implement that cache.

oleg-nenashev:
@bkmeneguello 
Sorry for the missing response. I'm quite busy with personal TODOs, so my responses may be seriously delayed. Don't hesitate to ping me

bkmeneguello:
@oleg-nenashev I've implemented the cache, a simple solution. In your opinion the cache parameters (max-size, ttl and concurrency) should be parameterized in a view? These entire feature should be an opt-in?
Thanks

bkmeneguello:
bump?

oleg-nenashev:
@bkmeneguello Thanks for the reminder. I was extremely busy with some personal things, but hopefully I'll be able to process incoming PRs on this NY break

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -53,6 +65,14 @@
   /** Map associating each {@link Role} with the concerned {@link User}s/groups. */
   private final SortedMap <Role,Set<String>> grantedRoles;
 
+  private static final Logger LOGGER = Logger.getLogger(RoleMap.class.getName());
+  
+  private final Cache<String, UserDetails> cache = CacheBuilder.newBuilder()
+          .softValues()
+          .maximumSize(100)
```
oleg-nenashev: Needs to be a global configuration/system property. 100 is a good default value, but I can imagine installation requiring more users

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -76,14 +96,31 @@ private boolean hasPermission(String sid, Permission p, RoleType roleType, Acces
...
                 }
             } // Default handling
             else {
                 return true;
             }
+        } else {
+            try {
+                UserDetails userDetails = cache.getIfPresent(sid);
+                if (userDetails == null) {
+                    userDetails = Jenkins.getInstance().getSecurityRealm().loadUserByUsername(sid);
```
oleg-nenashev: This method may throw UserMayOrMayNotExistException. I suspect it needs to be captured & logged

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -76,14 +96,31 @@ private boolean hasPermission(String sid, Permission p, RoleType roleType, Acces
...
+                }
+                for (GrantedAuthority grantedAuthority : userDetails.getAuthorities()) {
+                    if (grantedAuthority.getAuthority().equals(role.getName())) {
+                        return true;
+                    }
+                }
+            } catch (BadCredentialsException e) {
+                LOGGER.log(Level.FINE, "Bad credentials", e);
+            } catch (DataAccessException e) {
+                LOGGER.log(Level.FINE, "failed to access the data", e);
```
oleg-nenashev: Both exceptions smell like a fatal error. Or not?

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
bkmeneguello: Ok. But this value of 100 is the cache size. Just the 100 LRU entries will be cached.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
bkmeneguello: This exception is a BadCredentialsException, captured below

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
bkmeneguello: Well, maybe can we discuss about this. Since this is a new behaviour, on top of previous one, if the user hasn't the appropriate roles, I've logged the exception but kept the original behaviour.
What you propose with "fatal error"?

bkmeneguello:
@oleg-nenashev do you agree with my last replies?

oleg-nenashev:
@bkmeneguello 
Sorry, missed the responses. I hope to review it on this weekend

bkmeneguello:
Thanks!

bkmeneguello:
@oleg-nenashev pls, did you had time to review this PR?

oleg-nenashev:
@bkmeneguello 
Sorry for the delay. I've manually tested it today, looks good to me. I hope to find some time to finally make a release on the next week

bkmeneguello:
Thanks!

oleg-nenashev:
Looking for the issue to reference it during the merge. It exists AFAIK.
🐌

oleg-nenashev:
Caused performance regressions, hence the feature will be partially disabled in #18

bkmeneguello:
absolutely ok

PR #14, opened 2015-12-16
write config sorted into config.xml

kazesberger:
**problem:**
every time one saves the config page of this plugin, the resulting xml got randomly reordered speaking of sid and permission elements. this makes diffing messy in case one wants to have the config.xml in a git repo.
**solution:**
i made SortedSets (TreeSet) where needed without touching the actual persistence code.

### DIFF
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
@@ -54,14 +54,7 @@ import hudson.security.PermissionGroup;
 import hudson.security.SidACL;
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.SortedMap;
+import java.util.*;
 import javax.servlet.ServletException;
 
 import hudson.util.VersionNumber;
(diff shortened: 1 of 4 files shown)


KostyaSha:
:+1:

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -159,7 +152,7 @@ public ACL getACL(Computer computer) {
    * @param type The object type controlled by the {@link RoleMap}
    * @return All roles from the global {@link RoleMap}
    */
-  public SortedMap<Role, Set<String>> getGrantedRoles(String type) {
+  public SortedMap<Role, SortedSet<String>> getGrantedRoles(String type) {
```
oleg-nenashev: I would still return Sets, but OK

oleg-nenashev:
The change needs migration tests at least. Just to make sure that the upgrade works correctly

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
lanwen: +1 for Set. Don't mix interface and implementation

KostyaSha:
Wouldn't it be possible convert set to treeset only during deserialization?

darxriggs:
Is there someone still having this pull request in mind?

mawinter69:
Using TreeSet has also the disadvantage that is has worse performance compared to HashSet.
So a TreeSet should only be used when writing the config xml or when generating the CasC yaml

PR #18, opened 2016-06-12
[JENKINS-35515] - Fix the performance issue (regression in 2.3.0)

oleg-nenashev:
- [x] Allow tweaking cache settings.
- [x] Disable user authorities handling by default

This change should solve the reported performance issues.
https://issues.jenkins-ci.org/browse/JENKINS-35515

@reviewbybees @jenkinsci/code-reviewers @bkmeneguello

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/rolestrategy/Settings.java
+++ b/src/main/java/org/jenkinsci/plugins/rolestrategy/Settings.java
@@ -1,0 +1,74 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Oleg Nenashev.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/rolestrategy/Settings.java
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Enabling processing of User Authorities.
+     * Alters the behavior of 
+     * {@link RoleMap#hasPermission(java.lang.String, hudson.security.Permission, com.synopsys.arc.jenkins.plugins.rolestrategy.RoleType, hudson.security.AccessControlled)}.
+     * Since 2.3.0 this value was {@code true}, but it has been switched due to the performance reasons.
+     * The behavior can be reverted (even dynamically via System Groovy Script).
+     * @since 2.3.1
+     */
+    @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "We want to be it modifyable on the flight")
+    public static boolean THREAT_USER_AUTHORITIES_AS_ROLES =
```
batmat: @oleg-nenashev Didn't you mean _treat_ instead _threat_?

oleg-nenashev: yep :(

sschuberth:
Nit: The commit message say "deefault" instead of "default", and also the title is a bit long and should be split into body text.

oleg-nenashev:
@sschuberth It's not a problem, because the PR will be squashed. So the final commit message will be similar to the PR title

src/test/java/org/jenkinsci/plugins/rolestrategy/UserAuthoritiesAsRolesTest.java
```diff
@@ -24,6 +25,16 @@
 
     @Rule public JenkinsRule j = new JenkinsRule();
 
+    @Before
+    public void enableUserAuthorities() {
+        Settings.TREAT_USER_AUTHORITIES_AS_ROLES = true;
+    }
+    
+    @Before
```
oleg-nenashev: Typo: Should be "after"

varmenise:
LGTM 🐝

oleg-nenashev:
@reviewbybees done
@sschuberth Are you fine with the explanation?

sschuberth:
@oleg-nenashev Yes, if the PR will be squash the commit message obviously does not need to be fixed. However, I personally prefer to review "the real thing" and not squash but just merge that later.

oleg-nenashev:
I've switched to squashing after some backporting fun.
I used to do squashes before the final merge, but now GitHub saves some time by doing squash and rebase at the same time

PR #20, opened 2016-06-12
[JENKINS-35652] - Authorities resolution: Catch Runtime Exceptions from the SecurityRealm

oleg-nenashev:
This check suppresses runtime exceptions and prevents failure of the Authorization Strategy, which is fatal for Jenkins.

https://issues.jenkins-ci.org/browse/JENKINS-35652

### DIFF
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
@@ -119,6 +119,10 @@ public class RoleMap {
                 LOGGER.log(Level.FINE, "Bad credentials", e);
             } catch (DataAccessException e) {
                 LOGGER.log(Level.FINE, "failed to access the data", e);
+            } catch (RuntimeException ex) {
+                // There maybe issues in the logic, which lead to IllegalStateException in Acegi Security (JENKINS-35652)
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
CC @sschuberth

sschuberth:
Hmm. I'm not really feeling comfortable with swallowing this exception as I'm not sure about the security implications. While I tend to agree that the issue might be caused due to some interaction with another bug in the Active Directory Plugin, I know for a fact that downgrading only the Role Strategy Plugin from 2.3.0 to 2.2.0 makes the issue disappear. Thus I believe some more investigation into what change between the version causes this issue is required before an appropriate fix can be made.

oleg-nenashev:
@sschuberth 

> I'm not really feeling comfortable with swallowing this exception as I'm not sure about the security implications

We do not swallow it. The issue is being printed to the log, and we just return <code>false</code> from hasPermission(). 

> Thus I believe some more investigation into what change between the version causes this issue is required before an appropriate fix can be made.

 It is being caused by #13 . The entire authorities processing is a new logic added into 2.3.0. These routines will be disabled by default in any case (#18), but I want to make the behavior robust against the issue at least.

varmenise:
🐝

sschuberth:
@oleg-nenashev Ok, thanks for the explanation. The diff I was looking at was too narrow.

PR #21, opened 2016-08-14
[JENKINS-28975] Added support for View permissions at the View level rather than only at global level

pskumar448:
Added possibility to setup permissions at view level rather giving permissions at global level.

This will hide the View from top tab bar, but if the user try to access via direct URL it will be accessible.

New feature PR (https://github.com/jenkinsci/jenkins/pull/2499) to has been submitted to Jenkins Core to handle direct URL authorization of the user on view. 

https://issues.jenkins-ci.org/browse/JENKINS-28975

### DIFF
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
@@ -80,6 +80,7 @@ public class RoleBasedAuthorizationStrategy extends AuthorizationStrategy {
   public final static String GLOBAL    = "globalRoles";
   public final static String PROJECT   = "projectRoles";
   public final static String SLAVE     = "slaveRoles";
+  public final static String VIEW      = "viewRoles";
   public final static String MACRO_ROLE = "roleMacros";
   public final static String MACRO_USER  = "userMacros";
   
@@ -137,6 +138,23 @@ public class RoleBasedAuthorizationStrategy extends AuthorizationStrategy {
     @Override
     public ACL getACL(Computer computer) {
        return getACL(SLAVE, computer.getName(), RoleType.Slave, computer);
+    }
+
+    @Override
+    public ACL getACL(View view) {
+        return getACL(VIEW, getViewFullName(view), RoleType.View, view);
+    }
+
+    String getViewFullName(View view) {
+        StringBuilder sb = new StringBuilder();
+        sb.append(view.getOwnerItemGroup().getFullName());
+        if(sb.length()==0)  {
+            return view.getViewName();
+        }
+        else {
+            sb.append("/" + view.getViewName());
+            return sb.toString();
+        }
     }
   
   /**
@@ -449,7 +467,7 @@ public class RoleBasedAuthorizationStrategy extends AuthorizationStrategy {
 
       // If the form contains data, it means the method has been called by plugin
       // specifics forms, and we need to handle it.
-      if (formData.has(GLOBAL) && formData.has(PROJECT) && formData.has(SLAVE) && oldStrategy instanceof RoleBasedAuthorizationStrategy) {
+      if (formData.has(GLOBAL) && formData.has(PROJECT) && formData.has(SLAVE) && formData.has(VIEW) && oldStrategy instanceof RoleBasedAuthorizationStrategy) {
         strategy = new RoleBasedAuthorizationStrategy();
 
         JSONObject globalRoles = formData.getJSONObject(GLOBAL);
@@ -478,6 +496,7 @@ public class RoleBasedAuthorizationStrategy extends AuthorizationStrategy {
 
         ReadRoles(formData, PROJECT, strategy, (RoleBasedAuthorizationStrategy)oldStrategy);
         ReadRoles(formData, SLAVE, strategy, (RoleBasedAuthorizationStrategy)oldStrategy);
+        ReadRoles(formData, VIEW, strategy, (RoleBasedAuthorizationStrategy)oldStrategy);
       }
       // When called from Hudson Manage panel, but was already on a role-based strategy
       else if(oldStrategy instanceof RoleBasedAuthorizationStrategy) {
@@ -592,6 +611,10 @@ public class RoleBasedAuthorizationStrategy extends AuthorizationStrategy {
             groups.remove(PermissionGroup.get(SCM.class));
             groups.remove(PermissionGroup.get(Run.class));
         }
+        else if (type.equals(VIEW)) {
+            groups = new ArrayList<PermissionGroup>();
+            groups.add(PermissionGroup.get(View.class));
+        }
         else {
             groups = null;
         }
@@ -610,6 +633,9 @@ public class RoleBasedAuthorizationStrategy extends AuthorizationStrategy {
       }
       else if (type.equals(SLAVE)) {
           return p!=Computer.CREATE && p.getEnabled();
+      }
+      else if (type.equals(VIEW)) {
+          return p!=View.CREATE && p.getEnabled();
       }
       else {
         return false;
--- a/src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/assign-view-roles.jelly
+++ b/src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/assign-view-roles.jelly
@@ -1,0 +1,145 @@
+<!--
+  - The MIT License
+  -
+  - Copyright (c) 2016, Suresh.
+  -
+  - Original file: manage-project-roles.jelly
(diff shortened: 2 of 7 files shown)


src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/assign-view-roles.jelly
```diff
@@ -0,0 +1,145 @@
+<!--
+  - The MIT License
+  -
+  - Copyright (c) 2013, Oleg Nenashev, Synopsys Inc.
```
oleg-nenashev: Wrong license header

oleg-nenashev: Imho even the slave file needs some rework

pskumar448:
OK. I will update the license header and push the changes.

pskumar448:
@oleg-nenashev 
License headers for the files manage-view-roles.jelly and assign-view-roles.jelly updated and pushed changes.

pskumar448:
@oleg-nenashev 
Got any chance to look this PR.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -592,6 +599,10 @@ else if (type.equals(SLAVE)) {
             groups.remove(PermissionGroup.get(SCM.class));
             groups.remove(PermissionGroup.get(Run.class));
         }
+        else if (type.equals(VIEW)) {
+            groups = new ArrayList<PermissionGroup>();
+            groups.add(PermissionGroup.get(View.class));
```
oleg-nenashev: Blacklisting of groups above is made for a reason. If somebody adds a permission for view outside this group (e.g. MANAGE_OWNERSHIP.View), it won't appear in the table. For old plugins there is no way to add permissions to existing groups IIRC

oleg-nenashev:
Generally looks good to me. Need to perform some testing

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
pskumar448: If I get all the PermissionGroups then removing other except the View.class but Credentials section is coming in table so to eliminate that section which is not having any importance created a new list with just View PermissionGroup.

pskumar448:
@oleg-nenashev 
After testing this PR let me know any changes required.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
pskumar448: @oleg-nenashev 
Shall I do it like the previous implementation for VIEW too as you said that it has been implemented for a reason. 
Or else is this OK to add View.class only.

oleg-nenashev: Let's keep it as is by now

pskumar448: OK.

pskumar448:
@oleg-nenashev 
Any update on this.

oleg-nenashev:
I rarely do testing during the working week :(
Hope to spend some time on the weekend

oleg-nenashev:
The current permission check approach does not work correctly for Views in Folders Plugin. Seems you may have to emulate a kind of getFullName() method for views. And same for user custom views, for which the current behavior seems to be may be a security risk (needs core code review).

Also, I see the strange layout tabbing in Safari, which didn't use to happen before the patch (needs to be confirmed).

screen shot 2016-08-20 at 23 43 13

pskumar448:
@oleg-nenashev 
Ok. I will update with view full name for folder level views.

pskumar448:
@oleg-nenashev 
Emulated view full name to add support for folder level views. Let me know feedback on this after testing.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -138,6 +139,17 @@ public ACL getACL(AbstractItem project) {
     public ACL getACL(Computer computer) {
        return getACL(SLAVE, computer.getName(), RoleType.Slave, computer);
     }
+
+    @Override
+    public ACL getACL(View view) {
+        return getACL(VIEW, getViewFullName(view), RoleType.View, view);
+    }
+
+    String getViewFullName(View view) {
```
oleg-nenashev: This code will be invoked VERY frequently. It really makes sense to replace recursive call by a loop with StringBuilder

pskumar448:
Ok. I will modify the according to your feedback.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -138,6 +139,23 @@ public ACL getACL(AbstractItem project) {
...
+        StringBuilder sb = new StringBuilder();
+        sb.append(view.getOwnerItemGroup().getFullName());
+        if(sb.length()==0)  {
+            return view.getViewName();
+        }
+        else {
+            sb.append("/" + view.getViewName());
```
oleg-nenashev: The approach is not very safe in the case of slashes in the view name, which were allowed in old Jenkins versions, but it's a corner case we can ignore

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -138,6 +139,23 @@ public ACL getACL(AbstractItem project) {
...
     }
+
+    @Override
+    public ACL getACL(View view) {
+        return getACL(VIEW, getViewFullName(view), RoleType.View, view);
+    }
+
+    String getViewFullName(View view) {
+        StringBuilder sb = new StringBuilder();
+        sb.append(view.getOwnerItemGroup().getFullName());
```
oleg-nenashev: I'm not exactly sure what is going to happen for User private views here. Maybe it worth to create a couple of unit tests to address such cases. I can help with Role Strategy initialization framework for it

pskumar448: Ok. I will try to understand how User private views works as a functionality. Do you have any information regarding the crucial functionality of the User private views in Jenkins.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
pskumar448: We are using slash to separate the view name from the parent name. 
This functionality I have taken from the Jenkins core, referred some model classes how emulation of full name happening.

pskumar448:
Hi @oleg-nenashev 
Could you let me know regarding the User private views and any foresee issues with these changes in role-strategy plugin. I am clueless here regarding the User private views.

oleg-nenashev:
@pskumar448 
The issue is that one user with Jenkins.READ permission may be able to access personal views of another user if the role does not take users into account. E.g. "Foo" pattern may apply to "/Foo" and to "/user/oleg-nenashev/Foo" if you do not somehow resolve users

pskumar448:
OK. @oleg-nenashev 
Thanks for giving me the scenario, I will try to fix this issue.

pskumar448:
@oleg-nenashev 
I am working on this to fix the Views related to MyViews.

remm:
Hi guys, could you help me?
Looks like this [bug](https://issues.jenkins-ci.org/browse/JENKINS-39176) should be fixed with this pull request.

oleg-nenashev:
@remm Responded in the ticket

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
mawinter69, line 150: View names are ambiguous this way. When you use the nested views plugin There could be 2 views with that name
Assume you have a folder and in this you have 2 views: 
view1
view2
view2 is a nested view view, that contains also a view with name view1
folder/view1
         /view2/view1
For both views the above code would just return folder, which makes it unclear which is meant.

mawinter69:
no Intention to get this working.  full solution would require changes to core as well probably

PR #23, opened 2017-01-05
JENKINS-8075: Repeat the assign-project-roles header every 20 rows (GROUP NAME) and every 10 cols (USERNAME)

totoroliu:
JENKINS-8075: 
Repeat the assign-project-roles header every 20 rows (GROUP NAME) and every 10 cols (USERNAME)

### DIFF
--- a/src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/assign-project-roles.jelly
+++ b/src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/assign-project-roles.jelly
@@ -29,26 +29,49 @@
   <j:set var="id" value="${h.generateId()}"/>
 
   <table id="projectRoles" class="center-align global-matrix-authorization-strategy-table" name="data">
-
-      <!-- The first row will show grouping -->
-      <tr class="group-row">
-        <td class="start" />
-        <td class="pane-header blank">
-          ${%User/group}
-        </td>
-        <j:forEach var="role" items="${projectGrantedRoles}">
-          <td class="pane-header">
-            ${role.key.name}
-          </td>
-        </j:forEach>
-        <td class="stop" />
-      </tr>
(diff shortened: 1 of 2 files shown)


totoroliu:
jenkins - mozilla firefox_003

danechitoaie:
I think it would be better to have a fixed header from CSS. Like we have on job configuration page in Jenkins.

ljader:
For me this enhancement is a good start - of course there might be more "technically feasible" solution, but this is simple and effective.

@totoroliu is this fix playing nicely with "extenal addons" described here?:
https://wiki.jenkins-ci.org/display/JENKINS/Role+Strategy+Plugin

At least with "Jenkins Role Strategy UI enhancer" ?

oleg-nenashev:
I think it is fine as a temporary solution.
Will try to finally check and integrate it

dbrunkow:
What about clicking on a column to highlight it? That should be relatively easy.

dbrunkow:
Or click on the user and highlight their row

bhaberle:
@oleg-nenashev maybe a better solution is to have the top and/or side headers be floating?  I know that would help us out a ton.

bhaberle:
A sliding header or using the original pull request  (after being fixed) would be a short term solution.  @totoroliu ‘s solution is better and would be great  long term.

mawinter69:
There are now many improvements in the UI:
1. A Filter on the manage page allowing to filter the roles in global and item section
2. Filters for role and user on the assign roles page
3. permissions headers are rotated by 90° in the manage page
4. role headers are rotated by 90° on the assign page
5. tooltips on manage page: showing role, permissions and pattern
6. tooltips on assign page: showing role, user and pattern when hovering over a checkbox
   when hovering over the role header, the permissions for that role are shown

Hence I think that repeating the header every 20 rows is obsolete

bhaberle:
Agreed. Thank you!

Thank you,
Brett Haberle

PR #24, opened 2017-01-23
Increase horizontal spacing between columns

jbesta:
Increase spacing between columns to create wider visual gap and it doesn't appear that texts in neighbor columns "touch" each other.


oleg-nenashev:
I am not sure it is a good change for instances which have many permissions. IMHO it should be done via a custom CSS though it is possible to support multiple styles within the plugin

mawinter69:
With latest UI improvements, the permissions and roles are now rotated 90° and some little padding is added so the texts are not touching each other. This creates a very dense table.

PR #29, opened 2017-07-17
Create matrix-to-role-strategy.groovy

KostyaSha:
@oleg-nenashev just to not lost

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/role-strategy-plugin/29)


mawinter69:
Change would need to be adjusted for user/group separation

PR #31, opened 2017-08-21
Plugin facelift towards the 2.6.0 Release

oleg-nenashev:
- [x] - Update Parent POM
- [x] - New core baseline is 1.625.3, JTH hack is not required anymore
- [x] - FindBugs fixes in the code
- [x] - minor code Refactorings to reduce number of warnings in IDE

@reviewbybees @jenkinsci/code-reviewers just in case somebody is interested

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,12 +4,12 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.24</version>
+        <version>2.33</version>
     </parent>
     
     <artifactId>role-strategy</artifactId>
     <packaging>hpi</packaging>
-    <version>2.5.2-SNAPSHOT</version>
+    <version>2.6.0-SNAPSHOT</version>
     <name>Role-based Authorization Strategy</name>
     <url>http://wiki.jenkins-ci.org/display/JENKINS/Role+Strategy+Plugin</url>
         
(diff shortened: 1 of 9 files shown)


pom.xml
```diff
@@ -39,12 +39,9 @@
     </developers>
     
     <properties>
-        <!-- Minimal LTS baseline with https://issues.jenkins-ci.org/browse/JENKINS-18537 -->
-        <jenkins.version>1.554.3</jenkins.version>
-        <jenkins-test-harness.version>1.532.3</jenkins-test-harness.version>
-        <java.level>6</java.level>
+        <jenkins.version>1.625.3</jenkins.version>
```
daniel-beck: Surprisingly conservative given the usage stats.

Per http://stats.jenkins.io/pluginversions/role-strategy.html, 80% of 2.4.0 users are on Jenkins 2.7.1 or newer, and 90% of users of 2.5.x releases are on Jenkins 2.32.1 or newer.

oleg-nenashev: I was considering 1.651.3, but after some experiments it does not make much difference in terms of API. The only thing which I may need is user case sensitivity settings for https://issues.jenkins-ci.org/browse/JENKINS-20340 , but the core bump can be done in the pull request for that feature

daniel-beck: 2.19.x would get rid of the implied dependency on bouncycastle-api.

oleg-nenashev: Do not care much about it while the most of 10k+ installations plugins are based on 1.x branch.

oleg-nenashev:
@reviewbybees timeout

PR #32, opened 2017-09-18
update function of REST API

LegallGuillaume:
- doGetAllRoles
- doAssignRole
- doUnassignRole

### DIFF
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
@@ -330,9 +330,26 @@ public class RoleBasedAuthorizationStrategy extends AuthorizationStrategy {
     }
 
     /**
+    * Use by doAssignRole to add many sid from one query
+    *
+    * @param type (globalRoles, projectRoles, slaveRoles)
+    * @param roleName unassign role with sid
+    * @param sid  user ID to remove
+    */
+    private void assignRoleFn(String type, String roleName, String sid){
+      RoleMap roleMap = this.grantedRoles.get(type);
+        if (roleMap != null) {
+            Role role = roleMap.getRole(roleName);
+            if (role != null) {
+                assignRole(type, role, sid);
+            }
+        }
+    }
+
+    /**
      * API method to assign SID to role.
      * Example: {@code curl -X POST localhost:8080/role-strategy/strategy/assignRole --data "type=globalRoles&amp;roleName=ADM
-     * &amp;sid=username"}
+     * &amp;sid=username1,username2,group1"}
      *
      * @param type     (globalRoles, projectRoles, slaveRoles)
      * @param roleName name of role (single, no list)
@@ -346,15 +363,11 @@ public class RoleBasedAuthorizationStrategy extends AuthorizationStrategy {
                              @QueryParameter(required = true) String roleName,
                              @QueryParameter(required = true) String sid) throws IOException {
         checkAdminPerm();
-        RoleMap roleMap = this.grantedRoles.get(type);
-        if (roleMap != null) {
-            Role role = roleMap.getRole(roleName);
-
-            if (role != null) {
-                assignRole(type, role, sid);
-            }
-            persistChanges();
+        String[] sidlist = sid.split(',')
+        for (String sidOne : sidlist){
+          assignRoleFn(type, roleName, sidOne);
         }
+        persistChanges();
     }
 
     private static void persistChanges() throws IOException {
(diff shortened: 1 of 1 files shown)


LegallGuillaume:
@oleg-nenashev  hi :) , have you see this pull request ?

oleg-nenashev:
Yes, I have seen it. Just a bit snowed under the stuff

LegallGuillaume:
Oh great, I wait my turn :)

LegallGuillaume:
I was coding a API python for your plugin 
https://github.com/LegallGuillaume/jenkins-role-strategy-api-python

oleg-nenashev:
Nice :0 Do you plan to make it a part of the [python-jenkins](http://python-jenkins.readthedocs.io/en/latest/) library or to keep it independently? In any case, I am happy to add it to the list of 3rd-party tools in the new version of the plugin documentation (to be created sooner or later). Feel free to link it from the Wiki as well

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -429,9 +459,14 @@ public void doGetAllRoles(StaplerRequest req, StaplerResponse rsp) throws IOExce
         req.setCharacterEncoding("UTF-8");
         OutputStream os = rsp.getOutputStream();
         JSONObject json = new JSONObject();
-        RoleMap roleMap = this.grantedRoles.get(GLOBAL);
-        for (Map.Entry<Role, Set<String>> grantedRole : roleMap.getGrantedRoles().entrySet()) {
-          json.put(grantedRole.getKey().getName(), grantedRole.getValue());
+        type = req.getParameter("type")
```
oleg-nenashev: It could be simplified by using https://github.com/jenkinsci/role-strategy-plugin/blob/4d7e065d2d8fc569df6c51ed5698e50b9934977d/src/main/java/com/synopsys/arc/jenkins/plugins/rolestrategy/RoleType.java though it needs exception-free method. In the future. I am about adding support of typed APIs to the entire logic in order to speedup the API. Maybe I will create the new class in the `org/jenkinsci/plugins/rolestrategy` package to make it less confusing

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -429,9 +459,14 @@ public void doGetAllRoles(StaplerRequest req, StaplerResponse rsp) throws IOExce
...
+        if (type != null || (!type.equals(GLOBAL) && !type.equals(PROJECT) && !type.equals(SLAVE))){
+          RoleMap roleMap = this.grantedRoles.get(req.getParameter("type"));
+          for (Map.Entry<Role, Set<String>> grantedRole : roleMap.getGrantedRoles().entrySet()) {
+            json.put(grantedRole.getKey().getName(), grantedRole.getValue());
+          }
+        }else{
+          json.put("None", "None");
```
oleg-nenashev: It breaks the default behavior (retrieval of global roles). I would suggest to keep it if the parameter is not defined

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -429,9 +459,14 @@ public void doGetAllRoles(StaplerRequest req, StaplerResponse rsp) throws IOExce
...
         OutputStream os = rsp.getOutputStream();
         JSONObject json = new JSONObject();
-        RoleMap roleMap = this.grantedRoles.get(GLOBAL);
-        for (Map.Entry<Role, Set<String>> grantedRole : roleMap.getGrantedRoles().entrySet()) {
-          json.put(grantedRole.getKey().getName(), grantedRole.getValue());
+        type = req.getParameter("type")
+        if (type != null || (!type.equals(GLOBAL) && !type.equals(PROJECT) && !type.equals(SLAVE))){
+          RoleMap roleMap = this.grantedRoles.get(req.getParameter("type"));
+          for (Map.Entry<Role, Set<String>> grantedRole : roleMap.getGrantedRoles().entrySet()) {
+            json.put(grantedRole.getKey().getName(), grantedRole.getValue());
```
oleg-nenashev: In such case the pattern will not be reported. Maybe another data structure is required for that

oleg-nenashev: I would rather consider making the entire structure `@ExportedBean` and dumping all roles using Stapler. But it is another story

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
LegallGuillaume: Ok, I just understand some thing.

LegallGuillaume:
I think you can close this pull request.

I'll create a new pull request:
- function to return only json with: Role (in key) and his Permission (in value).
- function to update Role.

This is for the api python, to have Role with his permission and to update Role's permission


For my api-python (pyjars => PYthon Jenkins Api Role Strategy), I hope will make it a part of the python-jenkins library when the version pyjars version will be 1.0 (actually beta)

mtdeguzis:
Any update on this, work seems to have stopped. Is the REST API documented somewhere?

oleg-nenashev:
Not as much documented as I would like. I have started migrating all the docs to GitHub && added some docs there, but I have not finished it yet.

matpzu:
Hello @oleg-nenashev any news about documention of REST API ?

oleg-nenashev:
Nope :( Any contributions are welcome, my bandwidth is very limited and I rarely get to writing code or plugin docs on my own unfortunately

PR #34, opened 2017-10-04
[JENKINS-47265] - Stop requiring Matrix Auth flags to show Dangerous permissions

oleg-nenashev:
Just another proof that we need more integration testing :( The tests were working like a charm with Matrix Auth 1.0, but updating to 1.5+ reproduced the issue immediately.

@daniel-beck restricted the new incarnation of `GlobalMatrixAuthorizationStrategy.DescriptorImpl#showPermission(Permission)` method in Matrix Auth 2.0, so I decided to just stop calling the implementation there. Before 1.5 the method was just calling `p.enabled()` anyway: https://github.com/jenkinsci/matrix-auth-plugin/blob/matrix-auth-1.4/src/main/java/hudson/security/GlobalMatrixAuthorizationStrategy.java#L306

https://issues.jenkins-ci.org/browse/JENKINS-47265

@reviewbybees @daniel-beck

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.33</version>
+        <version>2.36</version>
     </parent>
     
     <artifactId>role-strategy</artifactId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -70,8 +70,14 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>matrix-auth</artifactId>
-            <version>1.0</version>
+            <version>1.7</version>
         </dependency>
+      <dependency>
+        <groupId>org.jenkins-ci.plugins</groupId>
+        <artifactId>cloudbees-folder</artifactId>
+        <version>5.18</version>
```
oleg-nenashev: It's not mandatory, but the tests are polluted by loading warnings otherwise

kuisathaverat:
LGTM 🐝

oleg-nenashev:
@reviewbybees done

shao123:
List roles in global roles with http://localhost:8080//role-strategy/strategy/getAllRoles
but what url can i use to know roles in project roles
image

oleg-nenashev:
https://github.com/jenkinsci/role-strategy-plugin/pull/32 is about doing it, but it needs changes. Please create an explicit JIRA ticket about missing functionality, AFAIK there is no such tickets right now

PR #35, opened 2017-10-04
Add Jenkinsfile

oleg-nenashev:
@reviewbybees


oleg-nenashev:
I doubt I should even request reviews for that.
@reviewbybees done

PR #37, opened 2018-02-27
Add optional Queryparameter to getAllRoles Service

marco2704:
According to improve this method without change its previous behavior, a new Queryparameter was added to getAllRoles method in RoleBasedAuthorizationStrategy. It may be useful in some case.

Thanks!

### DIFF
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
@@ -54,7 +54,6 @@ import javax.servlet.ServletException;
 
 import hudson.util.VersionNumber;
 
-import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -76,6 +75,7 @@ import org.jenkinsci.plugins.rolestrategy.permissions.PermissionHelper;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.QueryParameter;
+import org.kohsuke.stapler.Stapler;
 import org.kohsuke.stapler.StaplerRequest;
 import org.kohsuke.stapler.StaplerResponse;
 import org.kohsuke.stapler.interceptor.RequirePOST;
@@ -423,7 +423,7 @@ public class RoleBasedAuthorizationStrategy extends AuthorizationStrategy {
     }
 
     /**
-     * API method to get all groups/users with their role in globalRoles
+     * API method to get all groups/users with their role in any role type
      * Example: curl -X GET localhost:8080/role-strategy/strategy/getAllRoles?type=projectRoles
      *
      * @param type (globalRoles by default, projectRoles, slaveRoles)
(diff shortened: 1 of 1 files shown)


src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -424,25 +424,27 @@ public void doUnassignRole(@QueryParameter(required = true) String type,
...
     /**
      * API method to get all groups/users with their role in globalRoles
-     * Example: curl -X localhost:8080/role-strategy/strategy/getAllRoles
+     * Example: curl -X GET localhost:8080/role-strategy/strategy/getAllRoles?type=projectRoles
      *
-     * @throws IOException in case data fails
-     * @since 2.6.0
+     * @param type (globalRoles by default, projectRoles, slaveRoles)
+     *
+     * @since 2.8.0
```
oleg-nenashev: No need to modify that. The signature changed, but from Stapler PoV the methods are compatible

marco2704: Right, changed.

marco2704:
Hi! Another option with less changes:
```java 
/**
     * API method to get all groups/users with their role in globalRoles
     * Example: curl -X localhost:8080/role-strategy/strategy/getAllRoles?type=projectRoles
     *
     * @throws IOException in case data fails
     * @since 2.6.0
     */
    @Restricted(NoExternalUse.class)
    public void doGetAllRoles(StaplerRequest req, StaplerResponse rsp) throws IOException {
        checkAdminPerm();
        OutputStream os = rsp.getOutputStream();
        JSONObject json = new JSONObject();
        String type = req.hasParameter("type") ? req.getParameter("type") : GLOBAL;
        RoleMap roleMap = this.grantedRoles.get(type);
        for (Map.Entry<Role, Set<String>> grantedRole : roleMap.getGrantedRoles().entrySet()) {
          json.put(grantedRole.getKey().getName(), grantedRole.getValue());
        }
        os.write(json.toString().getBytes("UTF-8"));
    }
```

Both of them were tested in a staging Jenkins instance. Let me know any advice or improvement.

Thanks in advance!

oleg-nenashev:
I am fine with the current implementation. Generally it would be better to use https://github.com/stapler/stapler/blob/311dcfd92d229569e7aaaaae364f6e702188fddc/core/src/main/java/org/kohsuke/stapler/json/JsonHttpResponse.java so that the content type is set correctly.

@marco2704 Would it be possible to do so?

marco2704:
I'm not sure but i think it's not possible yet, because of the stapler version which seems to be 1.237 where json package does not exist.

marco2704:
Hi @oleg-nenashev! I've added a similar behavior, just until we can use [JsonHttpResponse.java](https://github.com/stapler/stapler/blob/311dcfd92d229569e7aaaaae364f6e702188fddc/core/src/main/java/org/kohsuke/stapler/json/JsonHttpResponse.java) , Regards!

marco2704:
Hi @oleg-nenashev any update about it? Hope we can merge it  😁

oleg-nenashev:
My bad, I missed the notifications. Merging/releasing

PR #39, opened 2018-05-17
Fix null pointer in addRole Service

PereBal:
When adding the **first** _projectRole_ through the addRole service, if _overwrite_ is set to true a NullPointerException is raised. Ie:
```sh
curl JENKINS_URL/role-strategy/strategy/addRole --data "type=projectRoles&roleName=ADM&permissionIds=hudson.model.Item.Read&overwrite=true"
```

This PR addresses that by adding a null check before trying to replace the previous role.


PbTG:
@oleg-nenashev Should I create an issue or something like that? (Apologies for my lack of knowledge regarding how to contribute to this project)

oleg-nenashev:
@PbTG Nope, this is fine. I just missed the notification somehow.  Releasing it

PR #43, opened 2018-08-13
[JENKINS-34545] - Support case insensitive user IDs in roles

seanmceligot:
This should fix JENKINS-34545 and related issues. It allows case insensitive matching to user/group values for Global and Item roles in Assign Roles.


seanmceligot:
OK, I'll look into that.

seanmceligot:
Here's a screenshot of what I have now. I was thinking Assign Roles screen is the most relevant place to put this but let me know what you think. Current it doesn't reload map without restarting Jenkins. I'm looking into that.

roles

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/Role.java
```diff
@@ -223,4 +223,11 @@ public boolean equals(Object obj) {
         }
         return true;
     }
+  public String toString() {
+        StringBuilder bldr = new StringBuilder("Role:");
+        bldr.append(name);
+        bldr.append(':');
+        bldr.append(pattern);
+        return bldr.toString();
```
KostyaSha: with String.format() you can do it in one line

oleg-nenashev:
Maybe it makes sense to move the setting to the common Authorization Strategy configuration screen (`config.jelly` in https://github.com/jenkinsci/role-strategy-plugin/tree/master/src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy) and making the setter a `@DataBoundSetter`. Otherwise this option will immediately become unmanageable from Configuration as Code.

No strong opinion tho, it is a wider problem than this PR

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -78,8 +78,10 @@
           .expireAfterWrite(Settings.USER_DETAILS_CACHE_EXPIRATION_TIME_SEC, TimeUnit.SECONDS)
           .build();
 
+  private final transient boolean caseInsensitiveUser;
```
oleg-nenashev: I am not sure what is the intention here. With the current logic the flag will be always false after the restart, and adding new roles from API won't behave as expected

daniel-beck:
Why not integrate with case sensitivity of the chosen security realm? Or what am I missing?

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
seanmceligot: The real configuration variable is in RoleBasedAuthorizationStrategy. That one is currently marshed/unmarshed explicitly in that RoleBasedAuthorizationStrategy.ConverterImpl. I just followed what was already there. Does the ConvererImpl entirely need to be converted or do you want the new variable managed elsewhere? 

This RoleMap.caseInsensitiveUser is a copy needed to create the TreeSet  when RoleMap.addRole is called. I made it transient because I didn't want it stored in config.xml and final to show that it was not the config.xml variable. It loads correctly at startup from config.xml, but I need to recreate all the TreeSets when it's changed realtime. 

config.xml
  ```<authorizationStrategy 
  class="com.michelin.cio.hudson.plugins.rolestrategy.RoleBasedAuthorizationStrategy">
      <caseInsensitiveUser>true</caseInsensitiveUser>
```

new variable placement on Configure Global Security page.

caseinsensitveconfiguresecurity

**Comment by ingwarsw** [comment]
@oleg-nenashev Any plans to integrate this..
Its really hard to explain users to login using always the same case..

**Comment by oleg-nenashev** [comment]
No short-term plans. If somebody wants to contribute, happy to share my
vision

**Comment by seanmceligot** [comment]
Just an update. My original fix was two lines lines of code.  

https://github.com/jenkinsci/role-strategy-plugin/compare/master...seanmceligot:case-insensitive-2.9.0
  
I started to add an configuration option turn it on and off as requested for legacy users, but my change wasn't compatible with configuration as code. I'd need to delve into the CASC framework if I wanted to continue.

**Comment by kiranraj365** [comment]
@seanmceligot  can we expect any progress on this issue in future.

**Comment by seanmceligot** [comment]
> @seanmceligot can we expect any progress on this issue in future.

Now should be a good time since configuration as code was merged. Let me look at adding the configuration parameter again and make sure it's CASC compatible.

**Comment by kiran1abc** [comment]
@seanmceligot thanks that helps us a lot.

**Comment by seanmceligot** [comment]
screenshot

**Comment by oleg-nenashev** [comment]
Thanks @seanmceligot ! It would be really great to get it over the line

File: src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java [hunk 3]
```diff
@@ -141,7 +158,7 @@ private boolean hasPermission(String sid, Permission permission, RoleType roleTy
     new RoleWalker() {
       public void perform(Role current) {
         if (current.hasAnyPermission(permissions)) {
-          if (grantedRoles.get(current).contains(sid)) {
+          if (hasSid( grantedRoles.get(current), userIdStrategy, sid))  {
```
AbhyudayaSharma: `contains()` in a set is a relatively fast operation. Now, you have to traverse the entire set to see if the role actually contains the sid.

The benchmarks suggest the same:
image
(The results were compared to the last successful build of master)

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -127,12 +141,15 @@ public RoleMap(@Nonnull SortedMap<Role,Set<String>> grantedRoles) {
   public static void invalidateDangerousPermissions() {
     PermissionHelper.DANGEROUS_PERMISSIONS.forEach(implyingPermissionCache::remove);
   }
-
+  private boolean hasSid(Set<String> roleAssignements, IdStrategy userIdStrategy, String sid) {
```
AbhyudayaSharma: Typo here: `roleAssignements` → `roleAssignments`

src/test/resources/org/jenkinsci/plugins/rolestrategy/Configuration-as-Code.yml
```diff
@@ -1,6 +1,7 @@
 jenkins:
   authorizationStrategy:
     roleBased:
+      userIdStrategyChoice: DEFAULT
```
AbhyudayaSharma: Shouldn't we have a test case where the `userIdStrategyChoice` is not present?

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
seanmceligot: Yes, I can't use Set because it's optimized at creation time time but IDStrategy as a configuration can be changed anytime.  I suppose I could use a case insensitive Set and then verify it using the correct IDStrategy at lookup time. Or, rebuild the Set every time the configuration changes, but that's a but trickier to get right.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
seanmceligot: fixed. Thank you. https://github.com/seanmceligot/role-strategy-plugin/commit/3be3e3bc8a41b3b713f07d13c03ef2cbb99cec31

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -116,13 +142,49 @@ public RoleBasedAuthorizationStrategy(Map<String, RoleMap> grantedRoles) {
...
+  }
+  @DataBoundSetter
+  public void setUserIdStrategyChoice(UserIdStrategyChoice userIdStrategyChoice) {
+      LOGGER.info("setUserIdStrategyChoice: "+ userIdStrategyChoice);
+      this.userIdStrategyChoice = userIdStrategyChoice;
+      globalRoles.invalidateCache();
+      agentRoles.invalidateCache();
+      itemRoles.invalidateCache();
+  }
+  public void setUserIdStrategyChoiceStr(String userIdStrategyChoice) {
```
oleg-nenashev: I suggest not making such public method. 

```suggestion
  @Restricted(NoExternalUse.class)
  public void setUserIdStrategyChoiceStr(String userIdStrategyChoice) {
```

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -116,13 +142,49 @@ public RoleBasedAuthorizationStrategy(Map<String, RoleMap> grantedRoles) {
...
+      LOGGER.info("setUserIdStrategyChoice"+ userIdStrategyChoice);
+      setUserIdStrategyChoice( UserIdStrategyChoice.valueOf(userIdStrategyChoice) );
```
oleg-nenashev: What if a non-existent value is passed. It will fail here, right?

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -680,6 +757,7 @@ public void doRolesSubmit(StaplerRequest req, StaplerResponse rsp) throws Unsupp
 
         req.setCharacterEncoding("UTF-8");
       JSONObject json = req.getSubmittedForm();
+      LOGGER.info("doRolesSubmit: "+json.toString());
```
oleg-nenashev: Unrelated. If you need it, please use `FINEST` level or lower

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -696,6 +774,7 @@ public void doAssignSubmit(StaplerRequest req, StaplerResponse rsp) throws Unsup
 
         req.setCharacterEncoding("UTF-8");
       JSONObject json = req.getSubmittedForm();
+      LOGGER.info("doAssignSubmit: "+json.toString());
```
oleg-nenashev: same

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
oleg-nenashev: Rebuilding set would be a good approach IMO.
I would be fine if the entire `RoleStrategy` object gets rebuilt on a configuration change

src/main/java/org/jenkinsci/plugins/rolestrategy/casc/RoleBasedAuthorizationStrategyConfigurator.java
```diff
@@ -33,6 +36,8 @@
 @Restricted({NoExternalUse.class})
 public class RoleBasedAuthorizationStrategyConfigurator extends BaseConfigurator<RoleBasedAuthorizationStrategy> {
 
+	  private static final Logger LOGGER = Logger.getLogger(RoleBasedAuthorizationStrategyConfigurator.class.getName());
```
oleg-nenashev: tabs vs. spaces

```suggestion
   private static final Logger LOGGER = Logger.getLogger(RoleBasedAuthorizationStrategyConfigurator.class.getName());  
```

src/test/java/org/jenkinsci/plugins/rolestrategy/RoleStrategyTest.java
```diff
@@ -7,8 +7,7 @@
 import hudson.model.FreeStyleProject;
 import hudson.model.Item;
 import hudson.model.User;
-import hudson.security.AuthorizationStrategy;
-import io.jenkins.plugins.casc.ConfigurationContext;
+import hudson.security.AuthorizationStrategy; import io.jenkins.plugins.casc.ConfigurationContext;
```
oleg-nenashev:
```suggestion
import hudson.security.AuthorizationStrategy;
import io.jenkins.plugins.casc.ConfigurationContext;
```

src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/assign-roles.jelly
```diff
@@ -56,7 +56,7 @@
               <td class="left-most">${title}</td>
               <j:forEach var="r" items="${it.strategy.getGrantedRoles(attrs.type)}">
                 <td width="*">
-                  <f:checkbox name="[${r.key.name}]" checked="${r.value.contains(attrs.sid)}"/>
+                    <f:checkbox name="[${r.key.name}]" checked="${r.value.contains(attrs.sid)}"/>
```
oleg-nenashev: Revert it 

```suggestion
                  <f:checkbox name="[${r.key.name}]" checked="${r.value.contains(attrs.sid)}"/>
```

src/test/resources/org/jenkinsci/plugins/rolestrategy/Configuration-as-Code.yml
oleg-nenashev: We should. Also, it is odd that all tests require "DEFAULT" and none really tests values set

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -127,12 +141,15 @@ public RoleMap(@Nonnull SortedMap<Role,Set<String>> grantedRoles) {
   public static void invalidateDangerousPermissions() {
     PermissionHelper.DANGEROUS_PERMISSIONS.forEach(implyingPermissionCache::remove);
   }
-
+  private boolean hasSid(Set<String> roleAssignments, IdStrategy userIdStrategy, String sid) {
+    return roleAssignments.stream().anyMatch( rolesid -> userIdStrategy.compare(rolesid, sid) == 0 );
```
AbhyudayaSharma: I was going through how Matrix Authorization plugin handles it. Looks to me that this check is incomplete. Since these sids can be group names too and there exists `SecurityRealm#getGroupIdStrategy()`, you will need to take care of that.

https://github.com/jenkinsci/matrix-auth-plugin/blob/8c7b7bcec67a03ab73498486fc2029b2b886541d/src/main/java/org/jenkinsci/plugins/matrixauth/AuthorizationContainer.java#L152

seanmceligot: I will look that at. Good catch.

zhaojinggang:
> This should fix JENKINS-34545 and related issues. It allows case insensitive matching to user/group values for Global and Item roles in Assign Roles.

Just check if there's any plan to release this feature?

PR #45, opened 2018-09-21
get role's sid

runzexia:


oleg-nenashev:
It was released in Role Strategy 2.10

PR #47, opened 2018-11-17
Change it.displayName to it.manageRolesName in manage-roles.jelly file

marco2704:
It was being displayed Manage and Assign Roles in Manage Roles page as title.


oleg-nenashev:
Thanks!

PR #48, opened 2018-12-01
[JENKINS-54900] improve concurrent security

runzexia:
In my usage scenario, I will insert a lot of roles. The order of insertion is slow for me.When I try to add a role concurrently, I get an exception.  
So I tried to use a concurrent safe container to store the role.And I packaged the plugin and tested the concurrent inserts in my environment.No exceptions continue to appear, and everything seems to be normal.  
Although replacing the container does not solve all the concurrency problems, I think it is effective for concurrent insertion of non-duplicate data.

### DIFF
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
@@ -66,6 +66,7 @@ import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import hudson.util.CopyOnWriteMap;
 import jenkins.model.Jenkins;
 import net.sf.json.JSONObject;
 import org.acegisecurity.acls.sid.PrincipalSid;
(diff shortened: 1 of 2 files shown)


runzexia:
override writeObject()

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -83,7 +84,7 @@
  * Role-based authorization strategy.
  * @author Thomas Maurel
  */
-public class RoleBasedAuthorizationStrategy extends AuthorizationStrategy {
+public class RoleBasedAuthorizationStrategy extends AuthorizationStrategy implements Serializable {
```
oleg-nenashev: does not need to be serializable

runzexia:
I asked in the dev mail list but I haven't replied yet, so I modified the PR according to my own thoughts. If there is time, can I help me review it?
1. Replace the thread-unsafe data structure so that the authentication operation can still be performed when the modification is made.
2. Use synchronized for thread synchronization for all write operations
3. And rewrite the writeObject method to get the latest state of the object, to ensure that serialization is saved.

runzexia:
I think rewriting readResolve() does not solve this problem, the problem occurs in the concurrency of write operations and iterative operations. Iterative operations are also performed during authentication. So replacing the data type should fix it

runzexia:
Hi @oleg-nenashev , do you think the current method can solve this problem? If you can solve it, merge the code?

oleg-nenashev:
Thanks @runzexia !

PR #49, opened 2018-12-26
Attempt to solve issue #45942

deepanshnagaria:
Proposing a fix to bug #45942 ("Bad error message/silent death with incorrectly setup permissions in the RoleBasedAuthorizationStrategy") by introducing a check on the permissions collected and prompting an error message if there's a problem.

### DIFF
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
@@ -526,9 +526,13 @@ public class RoleBasedAuthorizationStrategy extends AuthorizationStrategy {
 
               writer.startNode("permissions");
               for (Permission permission : role.getPermissions()) {
-                writer.startNode("permission");
-                writer.setValue(permission.getId());
-                writer.endNode();
+               if(permission!=null) {
+				  writer.startNode("permission");
+				  writer.setValue(permission.getId());
+				  writer.endNode();
+				}
+			   else
+			   	throw new NullPointerException("Cannot process the permissions as they have been incorrectly configured in the script");
               }
               writer.endNode();
 
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
@@ -531,8 +531,9 @@ public class RoleBasedAuthorizationStrategy extends AuthorizationStrategy {
 				  writer.setValue(permission.getId());
 				  writer.endNode();
 				}
-			   else
+			   else{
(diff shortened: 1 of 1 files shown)


src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -526,9 +526,14 @@ public void marshal(Object source, HierarchicalStreamWriter writer, MarshallingC
...
-                writer.startNode("permission");
-                writer.setValue(permission.getId());
-                writer.endNode();
+               if(permission!=null) {
+				  writer.startNode("permission");
+				  writer.setValue(permission.getId());
+				  writer.endNode();
+				}
+			   else{
+			   	throw new NullPointerException("Cannot process the permissions as they have been incorrectly configured in the script");
```
oleg-nenashev: I would rather recommend explicitly saying that `null` permission was discovered in the list of permission, preferable with a reference to the Role type and name.

deepanshnagaria: I understand the above stated concern and concur that the situation is possible. In my view, throwing an error earlier, when the list was being built ( line=295 )  would be better. I shall include the role name along with type as you mentioned. 

Please correct me if i am wrong or if any changes would improve the solution proposed.

Thanks.

oleg-nenashev:
Fixed the formatting + Improved the exception type and the text. Does it look good to you @deepanshnagaria ?

deepanshnagaria:
Looks fine to me but just to know was there any reason to switch to IOException? The compiler would have thrown a NPE in this case if we wouldn't have checked for the same. In my view if it was during the time of setting the permissions IOException would have made sense but for now while seeing a null permission I thought NPE should be thrown.
Please correct me if I am wrong.
Thanks.

oleg-nenashev:
> actually i am formatting the code as you may see in the screenshot but i an still figuring out why it is not being highlighted in the commit.

Tabs vs. Spaces, I'd guess

oleg-nenashev:
W.r.t the original question, I used `IOException`, because it is defined in the method interface. Using such checked exceptions requires API users to handle them, and I think it would be a positive improvement. But we can revert it or replace by `new IOException(new NullPointerException(...))` if you prefer

deepanshnagaria:
> > actually i am formatting the code as you may see in the screenshot but i an still figuring out why it is not being highlighted in the commit.
> 
> Tabs vs. Spaces, I'd guess

i got that :-)

deepanshnagaria:
> W.r.t the original question, I used `IOException`, because it is defined in the method interface. Using such checked exceptions requires API users to handle them, and I think it would be a positive improvement. But we can revert it or replace by `new IOException(new NullPointerException(...))` if you prefer

No sir i don't have any issues as such but was just asking the reason for the change. Please tell me if any further changes are required or if there's another more effective way for the same issue.

PR #50, opened 2019-01-02
[JENKINS-49102] make pattern display within quotes

vishaldotkhanna:


src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/manage-roles.jelly
```diff
@@ -49,7 +49,7 @@
               <td class="left-most">${title}</td>
               <j:if test="${!attrs.global}">
                 <td width="*" class="in-place-edit">
-                  ${h.escape(attrs.role.pattern.toString())}
+                  "${h.escape(attrs.role.pattern.toString())}"
```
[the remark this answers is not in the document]
vishaldotkhanna: @oleg-nenashev Makes sense. Made the changes.

oleg-nenashev:
@vishaldotkhanna I would appreciate if you could do unrelated changes in separate pull requests next time. It helps to reduce the review scope and to deliver changes separately if needed. It's ok for now, no need to rework

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -34,20 +34,14 @@
...
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.Map;
-import java.util.Set;
-import java.util.SortedMap;
-import java.util.SortedSet;
-import java.util.TreeMap;
-import java.util.TreeSet;
+import java.util.*;
```
oleg-nenashev: please do not use wildcard imports in the production code. It may lead to ambiguity in some cases. Automatic squashing can be disabled in your IDE

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -393,6 +387,25 @@ public void perform(Role current) {
     return roles;
   }
 
+    /**
+     * Get all job names matching the given pattern
+     * @param pattern Pattern to match against
+     * @return List of matching job names
+     */
+  public static List<String> getMatchingJobNames(Pattern pattern) {
+      List<Job> jobs = Jenkins.getInstance().getAllItems(Job.class);
```
oleg-nenashev: This approach is extremely greedy w.r.t memory. E.g. imagine you have 10000 jobs on the instance. We should limit the number of jobs to be fetched, ideally via various iterators or maybe even regexp predictors (e.g. extracting folder path before the first special regex semantic)

src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/manage-project-roles.jelly
```diff
@@ -176,8 +179,87 @@
...
       })();
 
+      var bindListenerToPattern = function(elem) {
+        elem.addEventListener('click', showMatchingProjects);
+      }
+
+      var showMatchingProjects = function() {
+        var pattern = this.text.substring(1, this.text.length - 1);    // Ignore quotes for the pattern
+
+        req = new XMLHttpRequest();
```
oleg-nenashev: I am not sure it is going to pass authentication in such way

vishaldotkhanna:
@oleg-nenashev That was my intention. I had created a new pull request but not sure why the new commits were added to the existing one. Will try to avoid this in the future.

src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/manage-roles.jelly
```diff
@@ -33,7 +33,16 @@
...
           <link rel="stylesheet" href="${rootURL}${app.VIEW_RESOURCE_PATH}/hudson/security/table.css" type="text/css" />
           <link rel="stylesheet" href="${rootURL}/plugin/role-strategy/css/role-strategy.css" type="text/css" />
+          <link rel="stylesheet" href="//code.jquery.com/ui/1.11.4/themes/smoothness/jquery-ui.css" />
           <script type="text/javascript" src="${rootURL}/plugin/role-strategy/js/table.js" />
+          <script type="text/javascript" src="http://code.jquery.com/jquery-latest.min.js" />
+          <script type="text/javascript">
+            <!--To avoid conflicts with other instances of jQuery in the namespace-->
+            $.noConflict();
+          </script>
+          <script src="https://code.jquery.com/ui/1.12.1/jquery-ui.min.js"
```
oleg-nenashev: If you need JQuery, use one from the plugin: https://plugins.jenkins.io/jquery

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -494,6 +495,22 @@ public void doGetAllRoles(@QueryParameter(fixEmpty = true) String type) throws I
...
+    /**
+     * API method to get a list of jobs matching a pattern
+     * Example: curl -X GET localhost:8080/role-strategy/strategy/getMatchingJobs?pattern=^staging.*
+     *
+     * @param pattern
+     * @throws IOException
+     */
+    @Restricted(NoExternalUse.class)
+    public void doGetMatchingJobs(@QueryParameter(required = true) String pattern) throws IOException {
+        List<String> matchingJobs = RoleMap.getMatchingJobNames(Pattern.compile(pattern));
```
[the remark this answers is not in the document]
vishaldotkhanna: @oleg-nenashev- I understand. Could you give a little more insight as to when to put an admin check in a method? I'm guessing here it is required since we are querying all the jobs in the instance.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
vishaldotkhanna: @oleg-nenashev Would something like this work better here? I'm not quite clear with the hierarchy between jobs, folders, projects etc.

`public static List<String> getMatchingJobNames(Pattern pattern, int maxJobs) {
      Iterator<Job> jobs = Jenkins.getInstance().allItems(Job.class).iterator();
      List<String> matchingJobNames = new ArrayList<>();
      maxJobs = Math.max(maxJobs, 0);    // An empty list will be returned for negative values of maxJobs

      while(jobs.hasNext() && matchingJobNames.size() < maxJobs) {
          Job job = jobs.next();
          Matcher m = pattern.matcher(job.getDisplayName());
          if(m.matches()) {
              matchingJobNames.add(job.getDisplayName());
          }
      }

      return matchingJobNames;
  }`

src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/manage-project-roles.jelly
vishaldotkhanna: @oleg-nenashev  I did not find an ajax request anywhere else in the code so wasn't entirely sure about this. Could you guide me about the authentication strategy used in the REST API?

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
oleg-nenashev: https://javadoc.jenkins.io/hudson/model/Items.html#allItems-hudson.model.ItemGroup-java.lang.Class- might be preferable. It is available since 2.37, but we can bump the plugin's core requirement to 2.60.3 if needed

src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/manage-project-roles.jelly
oleg-nenashev: Here is an example: https://github.com/jenkinsci/jenkins/blob/bdf9e4b389d8820911fb6feaa2118638af2724cd/war/src/main/webapp/scripts/hudson-behavior.js#L2878-L2909 . Generally Jenkins core should offer methods for that so that you do not need to worry

src/main/webapp/css/role-strategy.css
```diff
@@ -36,3 +36,12 @@
...
 }
+
+#matchingJobsForPattern ul {
+  list-style: none;
+}
+
+.matchingJobsDialog {
+  top: 50vh !important;
+  left: 50vw !important;
+}
```
pvtuan10: Better to add a new line at the end of the file

oleg-nenashev:
Sorry, I am a bit behind w.r.t reviews. I will try to take a look asap

pom.xml
```diff
@@ -83,6 +83,18 @@
...
+        <!-- https://mvnrepository.com/artifact/org.jenkins-ci.plugins/jquery -->
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>jquery</artifactId>
+            <version>1.11.2-0</version>
+        </dependency>
+        <!-- https://mvnrepository.com/artifact/org.jenkins-ci.plugins/jquery-ui -->
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>jquery-ui</artifactId>
```
oleg-nenashev: do you really use it?

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -393,6 +398,27 @@ public void perform(Role current) {
...
+     * @param maxJobs Max matching jobs to look for
+     * @return List of matching job names
+     */
+  public static List<String> getMatchingJobNames(Pattern pattern, int maxJobs) {
+      Iterator<Job> jobs = Items.allItems(Jenkins.getInstance(), Job.class).iterator();
+      List<String> matchingJobNames = new ArrayList<>();
+
+      while(jobs.hasNext() && matchingJobNames.size() < maxJobs) {
+          Job job = jobs.next();
+          Matcher m = pattern.matcher(job.getDisplayName());
```
oleg-nenashev: https://javadoc.jenkins.io/hudson/model/Item.html#getFullName-- should be used

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -393,6 +398,27 @@ public void perform(Role current) {
...
+          if(m.matches()) {
+              matchingJobNames.add(job.getDisplayName());
```
oleg-nenashev: Please avoid double method invocation. It may be heavy in some cases

src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/manage-roles.jelly
```diff
@@ -33,7 +33,23 @@
         <l:main-panel>
           <link rel="stylesheet" href="${rootURL}${app.VIEW_RESOURCE_PATH}/hudson/security/table.css" type="text/css" />
           <link rel="stylesheet" href="${rootURL}/plugin/role-strategy/css/role-strategy.css" type="text/css" />
+          <link rel="stylesheet" href="//code.jquery.com/ui/1.11.4/themes/smoothness/jquery-ui.css" />
```
oleg-nenashev: Does it use the plugin in such case? I am not sure

vishaldotkhanna: @oleg-nenashev I missed removing this from last time. I'm using the jQuery UI plugin included in the pom.xml file. I've added 1 more commit based on the review comments. Thanks a lot for your feedback.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
oleg-nenashev: @vishaldotkhanna Do you plan to address it?

pom.xml
daniel-beck: @oleg-nenashev This plugin is abandoned and looks like a terrible dependency to add. Last release _8 years ago_.

PR #54, opened 2019-01-28
[JENKINS-55803] - Attempt to solve issue:removing the intermediate collections created each time getRolesHavingPermission() is invoked

deepanshnagaria:
my attempt to fix the issue(55803,Performance: Replace RoleMap#getRolesHavingPermission() usages by iterators) by removing the intermediate collections each time the method is invoked. i did so by implementing the logic of the for loop in the roleWalker() itself. I removed the now not used method getRolesHavingPermission().

### DIFF
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
@@ -30,10 +30,12 @@ import com.synopsys.arc.jenkins.plugins.rolestrategy.Macro;
 import com.synopsys.arc.jenkins.plugins.rolestrategy.RoleMacroExtension;
 import com.synopsys.arc.jenkins.plugins.rolestrategy.RoleType;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import hudson.model.Items;
 import hudson.model.User;
 import hudson.security.AccessControlled;
 import hudson.security.Permission;
 import hudson.security.SidACL;
+import hudson.model.Job;
 import jenkins.model.Jenkins;
 import org.acegisecurity.BadCredentialsException;
 import org.acegisecurity.GrantedAuthority;
@@ -46,9 +48,11 @@ import org.springframework.dao.DataAccessException;
 
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.SortedMap;
@@ -61,6 +65,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * Class holding a map for each kind of {@link AccessControlled} object, associating
@@ -392,6 +397,28 @@ public class RoleMap {
     };
 
     return roles;
+  }
+  /**
+   * Get all job names matching the given pattern, viewable to the requesting user
+   * @param pattern Pattern to match against
+   * @param maxJobs Max matching jobs to look for
+   * @return List of matching job names
+   */
+  public static List<String> getMatchingJobNames(Pattern pattern, int maxJobs) {
+    Iterator<Job> jobs = Items.allItems(Jenkins.getInstance(), Job.class).iterator();
+    List<String> matchingJobNames = new ArrayList<>();
+
+    while(jobs.hasNext() && matchingJobNames.size() < maxJobs) {
+      Job job = jobs.next();
+      String jobName = job.getFullName();
+
+      Matcher m = pattern.matcher(jobName);
+      if(m.matches()) {
+        matchingJobNames.add(jobName);
+      }
+    }
+
+    return matchingJobNames;
   }
 
   /**
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
@@ -30,11 +30,14 @@ import com.synopsys.arc.jenkins.plugins.rolestrategy.Macro;
 import com.synopsys.arc.jenkins.plugins.rolestrategy.RoleMacroExtension;
 import com.synopsys.arc.jenkins.plugins.rolestrategy.RoleType;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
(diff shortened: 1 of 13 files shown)


deepanshnagaria:
Sorry for the issue in formatting i took reference from if else instead of this: (. I shall do that.  Sorry again for repeated issues in formatting,  i shall try to ignore further such issues.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -101,49 +101,66 @@ private boolean hasPermission(String sid, Permission p, RoleType roleType, Acces
...
+    final Permission per = p;
+    final boolean[] temp = {false};
+    // Get the implying permissions
+    for (; p!=null; p=p.impliedBy) {
+      permissions.add(p);
+    }
+    // Walk through the roles, and only add the roles having the given permission,
+    // or a permission implying the given permission
+    new RoleWalker() {
+      public void perform(Role current) {
```
runzexia: It would be better to add an overide annotation here.

deepanshnagaria: I don't think that would be correct, I think it would be correct if we were extending the abstract class **RoleWalker** . Don't you think?

deepanshnagaria:
@runzexia I myself thought about breaking the **RoleWalker** but with current implementation of it it doesn't seem possible. Do you want me to add the some methods or rewrite it to it, I would be happy to do that.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
deepanshnagaria: sure

deepanshnagaria:
@runzexia I myself thought of breaking the **RoleWalker**  but with current implementation it doesn't seem to be possible. But we can do the same by not calling the RoleWalker and implementing the logic of walk function, that could be done. Should I do that?

runzexia:
I think you can try to do that, I don't know how oleg thinks, this PR already looks good.

deepanshnagaria:
I shall do that in the next commit,  it won't take much i think. Thanks for the review.

oleg-nenashev:
@deepanshnagaria  @runzexia 
> I myself thought of breaking the RoleWalker but with current implementation it doesn't seem to be possible. But we can do the same by not calling the RoleWalker and implementing the logic of walk function, that could be done. Should I do that?

Yes, breaking the cycle will be an interesting change

RoleWalker is a part of the plugin codebase: https://github.com/jenkinsci/role-strategy-plugin/blob/529512ecfa80acdf5ae7684f03e1bfb58410e997/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java#L415-L436

My suggestion would be to create a follow-up ticket to support breaking the cycle from `RoleWalker` API (e.g. a new `abort()` method), and then this change can be landed in a follow-up PR.

deepanshnagaria:
@oleg-nenashev  I have created an issue https://issues.jenkins-ci.org/browse/JENKINS-55916 adding a point for memory, do tell if changes are required.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -101,49 +101,66 @@ private boolean hasPermission(String sid, Permission p, RoleType roleType, Acces
       /* if this is a dangerous permission, fall back to Administer unless we're in compat mode */
       p = Jenkins.ADMINISTER;
     }
-
-    for(Role role : getRolesHavingPermission(p)) {
-        
-        if(this.grantedRoles.get(role).contains(sid)) {
+    final Set<Permission> permissions = new HashSet<>();
+    final Permission per = p;
+    final boolean[] temp = {false};
```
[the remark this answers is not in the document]
oleg-nenashev: https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -101,54 +103,71 @@ private boolean hasPermission(String sid, Permission p, RoleType roleType, Acces
...
-                LOGGER.log(Level.FINE, "Bad credentials", e);
-            } catch (DataAccessException e) {
-                LOGGER.log(Level.FINE, "failed to access the data", e);
-            } catch (RuntimeException ex) {
-                // There maybe issues in the logic, which lead to IllegalStateException in Acegi Security (JENKINS-35652)
-                // So we want to ensure this method does not fail horribly in such case
-                LOGGER.log(Level.WARNING, "Unhandled exception during user authorities processing", ex);
+              }
+            }
+            catch (BadCredentialsException e) {
```
oleg-nenashev: Please revert such formatting changes.

Correct syntax here is...

```
} catch (BadCredentialsException e) {
```

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -417,16 +410,18 @@ protected Boolean hasPermission(Sid p, Permission permission) {
     RoleWalker() {
       walk();
     }
-
+    boolean abort=false;// should be set to true from within perform to break the loop
```
oleg-nenashev: It will work, but it would be better to introduce a new public setter so that external API users can also interrupt the execution

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
deepanshnagaria: Sorry i didn't note that

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
deepanshnagaria: Makes sense I will add the setter.  Thanks for the review.

deepanshnagaria:
> @deepanshnagaria there is a minor merge conflict after integration of other PRs. Are you able to resolve it? Or do you need some assistance?

yeah i shall do resolve that.

deepanshnagaria:
ignore last commit pls

oleg-nenashev:
@deepanshnagaria could you please configure your IDE to disable automatic code reformatting? Even if the code formatting is not good, massive formatting changes usually leads to merge conflicts. It is better to handle it in a separate PR next time

deepanshnagaria:
> LGTM apart from the code formatting

I have solved most of the formatting issues. Thanks for the review. Think, that it is ready for merge. Suggest changes if needed. @oleg-nenashev @runzexia

PR #56, opened 2019-02-01
[JENKINS-55916] - Attempt to solve issue

deepanshnagaria:
I propose a fairly simple approach of both the problems listed: A class variable **abort** which when set to true terminates the loop and making the collection **roles** as global variable of outer class solves the issue of it being created on every call of **walk()**.

### DIFF
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
@@ -55,6 +55,8 @@ import java.util.SortedMap;
 import java.util.SortedSet;
 import java.util.TreeMap;
 import java.util.TreeSet;
+import java.util.concurrent.ConcurrentSkipListMap;
+import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -418,15 +420,18 @@ protected Boolean hasPermission(Sid p, Permission permission) {
       walk();
     }
 
+    boolean abort=false;// should be set to true from within perform to break the loop
```
oleg-nenashev: not used now

deepanshnagaria:
Sir my intent was not to focus on the formatting but on the added 6 lines, the issue as noted 55916 and the explanation is only for those 6 lines of code only....

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
deepanshnagaria: That is what I propose to add sir,  that is what solves the issue, don't you think. I added this for breaking of loop conditionally in the Rolewalker.

PR #59, opened 2019-04-01
JENKINS-56834 Implement getACL for Node

LuisPiedra:


runzexia:
Hi Can you say why you want to add a new getACL method? Node seems to be more than just  slaves but also  master.

oleg-nenashev:
Sorry, I missed this PR

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -152,6 +153,12 @@ public ACL getACL(AbstractItem project) {
       return getACL(PROJECT, project.getFullName(), RoleType.Project, project);
     }
 
+
+    @Override
+    public ACL getACL(Node node) {
+       return getACL(SLAVE, node.getNodeName(), RoleType.Slave, node);
```
oleg-nenashev: Note that `node.getNodeName()` and `computer.getName()` return different values for the Master node. So it may have some unexpected results

rob68: I was having the same issue and need this change.  @oleg-nenashev I believe I was able to get around your issue with the following code:

`
    @Override
    @Nonnull
    public ACL getACL(@Nonnull Node node) {
        return getACL(node.toComputer());
    }
`

oleg-nenashev:
Yes, we definitely need to land it @rob68 . Will take a look

rexbut:
Can you validate this pull request?

shywei:
> Yes, we definitely need to land it @rob68 . Will take a look

Can we use 'computer.getDisplayName()' instead of 'computer.getName()'? 
```
@Override
@Nonnull
public ACL getACL(@Nonnull Computer computer) {
    return agentRoles.newMatchingRoleMap(computer.getDisplayName()).getACL(RoleType.Slave, computer)
            .newInheritingACL(getRootACL());
}

@Override
@Nonnull
public ACL getACL(@Nonnull Node node) {
    return getACL(node.toComputer());
}
```
With this we can use 'master' to match the master node, but the build permission still not work. After studying the core code of jenkins, I think the problem is that the 'getACL()' method in the 'Jenkins.java'.
```
@Override
public ACL getACL() {
    return authorizationStrategy.getRootACL();
}
```
Change the method to below, the problem seems to be resolved. But I'm not sure if this modification will cause other unknown problems. Any suggestion?
```
@Override
public ACL getACL() {
    return authorizationStrategy.getACL(this) == null ? authorizationStrategy.getRootACL():authorizationStrategy.getACL(this);
}
```
In addition to the code modification above , I also installed the 'Authorize Project' plugin and set 'Access Control for Builds' to 'Run as User who Triggered Build' in 'Configure Global Security'. In this way, I can control the tasks manually performed by the user, but there is still a problem when tasks were started by a trigger. The 'Job Restrictions Plugin' maybe a supplement but I still prefer a lack-permission message instead of skip the node.

PR #62, opened 2019-05-13
[JENKINS-57422] Slowdown config cleanup

AbhyudayaSharma:
Cleanup of the files used for the configuration to visually see the slowdown of the plugin.


AbhyudayaSharma:
This configuration is still not perfect.

```text
SEVERE: Failed Loading plugin Pipeline: Declarative Extension Points API v1.3.8 (pipeline-model-extensions)
java.io.IOException: Pipeline: Declarative Extension Points API v1.3.8 failed to load.
 - Pipeline: Groovy v2.59 is older than required. To fix, install v2.61.1 or later.
        at hudson.PluginWrapper.resolvePluginDependencies(PluginWrapper.java:652)
```
I couldn't figure out how to update the groovy version. 

I've also removed the `DEV_HOST`. I don't know if it is needed right now.

src/test/performance/jenkins2.sh
```diff
@@ -9,16 +9,9 @@ extra_java_opts=( \
...
-  echo "WARNING: DEV_HOST is undefined, localhost will be used. Some logic like Docker Cloud may work incorrectly."
-else
-  extra_java_opts+=( "-Dio.jenkins.dev.host=${DEV_HOST}" )
-fi
-
 if [[ "$DEBUG" ]] ; then
   extra_java_opts+=( \
-    '-Xdebug' \
-    '-Xrunjdwp:server=y,transport=dt_socket,address=5005,suspend=y' \
+    '-Xdebug'
```
oleg-nenashev: Hmm, is it going to work after that?

AbhyudayaSharma: There is no need to check if the `DEBUG` variable is set as the script calls jenkins.sh in the parent Dockerfile which performs exactly the same thing.

PR #63, opened 2019-05-20
[JENKINS-57423] Integrate JMH with the plugin

AbhyudayaSharma:
Enable running JMH benchmarks plugins with a Jenkins instance running in the forked VM. The instance is started similar to how [JenkinsRule](https://javadoc.jenkins.io/component/jenkins-test-harness/org/jvnet/hudson/test/JenkinsRule.html) does it. The instance is terminated after the benchmark is complete.

A part of the Google Summer of Code 2019 project to improve the performance of this plugin.

@oleg-nenashev @runzexia @Supun94


pom.xml
```diff
@@ -93,7 +93,46 @@
             <artifactId>jquery-ui</artifactId>
             <version>1.0.1</version>
         </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-core</artifactId>
+            <version>1.21</version>
```
oleg-nenashev: It makes JMH bundled to the final plugin distribution. the scope should be test

```suggestion
            <version>1.21</version>
            <scope>test</scope>
```

pom.xml
```diff
@@ -93,7 +93,46 @@
...
+        </dependency>
     </dependencies>
-</project>  
-  
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
```
oleg-nenashev: Is it coming from https://gist.github.com/nicoulaj/5517176 ? If so, a proper attribution would be nice

pom.xml
```diff
@@ -93,7 +93,46 @@
...
         </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-core</artifactId>
+            <version>1.21</version>
+        </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-generator-annprocess</artifactId>
+            <version>1.21</version>
```
oleg-nenashev:
```suggestion
            <version>1.21</version>
            <scope>test</scope>
```

pom.xml
AbhyudayaSharma: I've removed it as it was causing the build to fail. The JMH benchmarks are run from a JUnit test itself.

AbhyudayaSharma:
> Also, it is generally a good practice to reference JIRA tickets and to put some explanation to the ticket

I'm really sorry. I forgot about it.

src/test/java/jmh/PermissionBenchmark.java
```diff
@@ -0,0 +1,56 @@
...
+            Set<String> permissionSet = Collections.singleton("hudson.model.Hudson.Administer");
+            Role role = new Role("USERS", ".*", permissionSet, "description");
+            RoleMap roleMap = new RoleMap(new TreeMap<>( // expects a sorted map
+                    Collections.singletonMap(role, Collections.singleton("alice"))
+            ));
+
+            jenkins.setAuthorizationStrategy(new RoleBasedAuthorizationStrategy(
+                    Collections.singletonMap(RoleBasedAuthorizationStrategy.GLOBAL, roleMap)));
+
+            jenkins.setSecurityRealm(new JenkinsRule().createDummySecurityRealm());
```
AbhyudayaSharma: This seems to work for now

src/test/java/jmh/PermissionBenchmark.java
```diff
@@ -0,0 +1,56 @@
...
+
+public class PermissionBenchmark {
+    @State(Scope.Benchmark)
+    public static class JenkinsState extends JmhBenchmarkState {
+        private Authentication originalAuthentication = null;
+        private SecurityContext securityContext = null;
+
+        @Override
+        public void setup(Jenkins jenkins) {
+            Set<String> permissionSet = Collections.singleton("hudson.model.Hudson.Administer");
```
AbhyudayaSharma: I've taken this configuration from the respective [`config.xml`](https://github.com/jenkinsci/role-strategy-plugin/blob/master/src/test/resources/org/jenkinsci/plugins/rolestrategy/RoleAssignmentTest/testRoleAssignment/config.xml).

src/test/java/jmh/JmhBenchmarkState.java
```diff
@@ -0,0 +1,216 @@
+package jmh;
+
+import hudson.WebAppMain;
+import hudson.model.DownloadService;
+import hudson.model.Hudson;
+import hudson.model.JDK;
```
oleg-nenashev: Better to avoid unused imports

pom.xml
```diff
@@ -93,7 +93,38 @@
...
     </dependencies>
-</project>  
-  
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
```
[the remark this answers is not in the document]
AbhyudayaSharma: From [this](https://stackoverflow.com/questions/6178583/maven-does-not-find-junit-tests-to-run) Stack Overflow answer, surefire only finds methods that either start or end with `Test`. If you like I can rename it to something like `BenchmarkRunnerTest` but the named in my opinion doesn't make much sense.

Also, the plugin pom defines the following (as of now):
```xml
          <systemProperties>
            <property>
              <name>hudson.udp</name>
              <value>-1</value>
            </property>
            <property>
              <name>java.io.tmpdir</name>
              <value>${surefireTempDir}</value>
            </property>
          </systemProperties>
          <runOrder>alphabetical</runOrder>
        </configuration>
```
None of these seem to be necessary for now. If you like, I can copy this into the pom.xml as well.

src/test/java/jmh/JmhBenchmarkState.java
```diff
@@ -0,0 +1,216 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+@State(Scope.Benchmark)
+public abstract class JmhBenchmarkState implements RootAction {
+    private static final Logger LOGGER = Logger.getLogger(JmhBenchmarkState.class.getName());
+    private static final String contextPath = "/jenkins";
+
+    private final File baseTempDirectory = new File(System.getProperty("java.io.tmpdir"));
+    private final boolean withoutSpace = Boolean.getBoolean("jenkins.test.noSpaceInTmpDirs");
```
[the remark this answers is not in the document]
AbhyudayaSharma: It's from Jenkins-test-harness. According to the [README](https://github.com/jenkinsci/jenkins-test-harness/blob/master/README.md#245-2019-jan-10):
> Temporary directories now by default have a space in their path, to flush out path handling mistakes. (The system property jenkins.test.noSpaceInTmpDirs=true suppresses this.)

I will remove this if you like.

src/test/java/jmh/JmhBenchmarkState.java
AbhyudayaSharma: These are all used in the file to configure up the instance that will be used for JMH benchmarks

src/test/java/jmh/JmhBenchmarkState.java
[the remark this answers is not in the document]
oleg-nenashev: Yes, I do not think we need it for performance tests. K.I.S.S.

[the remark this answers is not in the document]
AbhyudayaSharma: Done

AbhyudayaSharma:
It seems as if the build is timing out.

oleg-nenashev:
It is. Also, benchmarks now run on 4 platforms now. Maybe we could update Jenkinsfile to not run performance  as a part of the standard `buildPlugin()` step but to run it as an additional step after it. In order to do so, we would need a property flag in Java which would allow disabling performance tests

pom.xml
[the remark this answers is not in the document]
oleg-nenashev: i would suggest renaming the class for the sake of simplicity

PR #65, opened 2019-05-27
[JENKINS-57425] Run benchmarks only when a special Maven profile is defined

AbhyudayaSharma:
Run benchmarks separately from the main build
Follow-up to #63

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,8 +1,11 @@
 // Builds a module using https://github.com/jenkins-infra/pipeline-library
 buildPlugin(jenkinsVersions: [null, '2.150.2'])
-node {
-    stage('benchmark') {
-        List<String> mvnOptions = ['test', '-Dbenchmark']
-        infra.runMaven(mvnOptions)
-    }
-}
+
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -1,2 +1,4 @@
 // Builds a module using https://github.com/jenkins-infra/pipeline-library
 buildPlugin(jenkinsVersions: [null, '2.150.2'])
+List<String> mvnOptions = ['test', '-Dbenchmark']
+infra.runMaven(mvnOptions)
```
oleg-nenashev: Have you tested it locally? I do not think it is going to work. No node selection, for starters

AbhyudayaSharma: No, I did not test it locally. I thought the pull request build would verify it.

casz:
Please, these tests are incredibly slow. :scream:

casz:
I would prefer if these tests were not run on pull requests unless requested, as a contributor, I would hate to sit and wait for my incremental build or waiting for tests to pass.

You could use a declarative pipeline to use a when block.

PR #67, opened 2019-05-28
Add a benchmark for RoleMap#hasPermission

AbhyudayaSharma:
Adds a benchmark for `RoleMap#hasPermission`.

Benchmark Reports: [jmh-report.txt](https://github.com/jenkinsci/role-strategy-plugin/files/3227535/jmh-report.txt) (JSON file; GitHub doesn't allow .json extension)

Takes much longer with increased number of roles.

@jenkinsci/gsoc2019-role-strategy

### DIFF
--- a/src/test/java/jmh/BenchmarkRunner.java
+++ b/src/test/java/jmh/BenchmarkRunner.java
@@ -2,6 +2,7 @@ package jmh;
 
 import jmh.benchmarks.CascBenchmark;
 import jmh.benchmarks.PermissionBenchmark;
+import jmh.benchmarks.RoleMapBenchmark;
 import org.junit.Test;
 import org.openjdk.jmh.annotations.Mode;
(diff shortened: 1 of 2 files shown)


src/test/java/jmh/BenchmarkRunner.java
```diff
@@ -17,6 +18,7 @@ public void runJmhBenchmarks() throws Exception {
         Options options = new OptionsBuilder()
                 .include(PermissionBenchmark.class.getName() + ".*")
                 .include(CascBenchmark.class.getName() + ".*")
+                .include(RoleMapBenchmark.class.getName() + ".*")
```
oleg-nenashev: From a framework standpoint, it would be cool to somehow automate discovery of the Benchmark classes

AbhyudayaSharma: I implemented it! Please see #69

PR #70, opened 2019-05-29
Add hasPermission benchmarks for 100, 200 and 500 roles

AbhyudayaSharma:
Adds benchmarks for multiple roles as requested by @runzexia .


oleg-nenashev:
Release drafter also supports the `test` tag: https://github.com/jenkinsci/.github/blob/master/.github/release-drafter.yml . If we have many such changes going forward, we could consider this category for the changelogs

PR #72, opened 2019-05-31
Add a benchmark using a Malicious Regex

AbhyudayaSharma:
This malicious regex takes a lot of time to evaluate. This is what I had used for the Docker configuration in #62 . Now each operation takes ~ 100 ms.

### DIFF
--- a/src/test/java/jmh/benchmarks/MaliciousRegexBenchmark.java
+++ b/src/test/java/jmh/benchmarks/MaliciousRegexBenchmark.java
@@ -1,0 +1,59 @@
+package jmh.benchmarks;
+
+import com.cloudbees.hudson.plugins.folder.Folder;
+import com.michelin.cio.hudson.plugins.rolestrategy.RoleBasedAuthorizationStrategy;
+import hudson.model.User;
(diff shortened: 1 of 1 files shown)


src/test/java/jmh/benchmarks/MaliciousRegexBenchmark.java
```diff
@@ -0,0 +1,59 @@
...
+            for (int i = 0; i < 200; i++) {
+                jenkins.createProject(Folder.class, "Foooooolder" + i);
+            }
+
+            Random rand = new Random(33);
+            for (int i = 0; i < 300; i++) {
+                if (rand.nextBoolean()) {
+                    rbas.doAddRole(RoleBasedAuthorizationStrategy.PROJECT, "role" + i,
+                            "hudson.model.Item.Discover,hudson.model.Item.Read,hudson.model.Item.Build",
+                            "true", "F(o+)+lder[" + rand.nextInt(10) + rand.nextInt(10) + "]{1,2}");
```
AbhyudayaSharma: This is the regex causing the slowdown.

PR #73, opened 2019-05-31
Add support of WebClient-based tests for REST API and Web UI benchmarks

AbhyudayaSharma:
Adds support for running WebClient in JMH benchmarks.

Draft until jQuery from #50 is 404 from the following test:
```java
public class RandomTest {
    @Rule
    public JenkinsRule j = new JenkinsRule();

    @Test
    public void test() throws Exception {
        JenkinsRule.WebClient webClient = j.createWebClient();
        webClient.login("user34", "user34");
        webClient.goTo("");
    }
}
```
@jenkinsci/gsoc2019-role-strategy

### DIFF
--- a/src/test/java/jmh/benchmarks/WebClientBenchmark.java
+++ b/src/test/java/jmh/benchmarks/WebClientBenchmark.java
@@ -1,0 +1,37 @@
+package jmh.benchmarks;
+
+import jmh.JmhBenchmarkState;
+import jmh.JmhJenkinsRule;
+import org.jvnet.hudson.test.JenkinsRule;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.infra.Blackhole;
+
+public class WebClientBenchmark {
+    public static class JenkinsState extends JmhBenchmarkState {
+        JenkinsRule.WebClient webClient = null;
+
+        @Override
+        public void tearDown() {
+            webClient.close();
+        }
+
+        @Override
+        public void setup() {
+            JmhJenkinsRule jenkinsRule = new JmhJenkinsRule();
+            webClient = jenkinsRule.createWebClient();
+        }
+
+        @Setup(Level.Iteration)
+        public void login() throws Exception {
+            webClient.login("user33", "user33");
+        }
+    }
+
+    @Benchmark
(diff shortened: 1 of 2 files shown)


src/test/java/jmh/benchmarks/WebClientBenchmark.java
```diff
@@ -1,37 +1,48 @@
...
-        @Override
-        public void tearDown() {
-            webClient.close();
-        }
+        @Setup(Level.Iteration)
+        public void setup() throws Exception {
+            // JQuery UI is 404 from these tests so disable stopping benchmark when it is used.
+            JmhJenkinsRule j = new JmhJenkinsRule();
+            webClient = j.createWebClient();
+            webClient.setJavaScriptEnabled(false); // have to disable JavaScript since it cannot find jQuery
```
AbhyudayaSharma: This may not represent the actual page load times as we are ignoring the AJAX build queue.

oleg-nenashev: I suggest putting it in a TODO comment so that this concern is represented in the code.
Anyway, I do not think it is critical for this story. Later we can implement a separate benchmark for the async logic

AbhyudayaSharma: Done!

src/test/java/jmh/benchmarks/WebClientBenchmark.java
```diff
@@ -0,0 +1,48 @@
...
+        @Setup(Level.Iteration)
+        public void setup() throws Exception {
+            // JQuery UI is 404 from these tests so disable stopping benchmark when it is used.
+            JmhJenkinsRule j = new JmhJenkinsRule();
+            webClient = j.createWebClient();
+            webClient.setJavaScriptEnabled(false); // TODO enable JavaScript when we can find jQuery
+            webClient.setThrowExceptionOnFailingStatusCode(false);
+            webClient.getOptions().setPrintContentOnFailingStatusCode(false); // reduce 404 noise
+
+            webClient.login("user33", "user33");
```
oleg-nenashev: If it is a mock user, better to call it "mockUser" or so so that there is no assumptions it is something real

oleg-nenashev:
> It looks like some setup code could be moved to a utility method in the framework. Can be done later once the bigger part of JTH is integrated.

NVM, I forgot again about the Jquery UI discussion

PR #74, opened 2019-06-02
[JENKINS-34337] - Ensure that the Role-Based naming strategy is applied correctly during renames

nightowlengineer:
Very similar to #16 but with the performance improvements requested in the comment chain.

This change fixes [JENKINS-34337](https://issues.jenkins-ci.org/browse/JENKINS-34337) "Job Naming Strategy doesn't enforce restriction on rename" (it also appeared to not enforce it during item creation resulting in a job that could be created but then not configured/renamed etc).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/rolestrategy/RoleBasedProjectNamingStrategy.java
+++ b/src/main/java/org/jenkinsci/plugins/rolestrategy/RoleBasedProjectNamingStrategy.java
@@ -5,6 +5,7 @@ import com.google.common.cache.CacheBuilder;
 import com.michelin.cio.hudson.plugins.rolestrategy.Messages;
 import com.michelin.cio.hudson.plugins.rolestrategy.Role;
 import com.michelin.cio.hudson.plugins.rolestrategy.RoleBasedAuthorizationStrategy;
+import com.synopsys.arc.jenkins.plugins.rolestrategy.RoleType;
 import hudson.Extension;
 import hudson.model.Failure;
 import hudson.model.Item;
@@ -50,13 +51,13 @@ public class RoleBasedProjectNamingStrategy extends ProjectNamingStrategy implem
     public void checkName(String name) throws Failure {
         boolean matches = false;
         ArrayList<String> badList = null;
-        final AuthorizationStrategy auth = Jenkins.getInstance().getAuthorizationStrategy();
+        final AuthorizationStrategy auth = Jenkins.get().getAuthorizationStrategy();
         if (auth instanceof RoleBasedAuthorizationStrategy) {
             final RoleBasedAuthorizationStrategy rbas = (RoleBasedAuthorizationStrategy) auth;
 
             // Check whether the user has a global role allowing them to carry out this action
             boolean globalRolePermitted = false;
-            final SortedMap<Role, Set<String>> gRole = rbas.getGrantedRoles(RoleBasedAuthorizationStrategy.GLOBAL);
+            final SortedMap<Role, Set<String>> gRole = rbas.getGrantedRoles(RoleType.Global);
             for (final SortedMap.Entry<Role, Set<String>> entry : gRole.entrySet()) {
                 if (entry.getKey().hasPermission(Item.CREATE)) {
                     globalRolePermitted = true;
@@ -69,7 +70,7 @@ public class RoleBasedProjectNamingStrategy extends ProjectNamingStrategy implem
             }
 
             // Check whether there's a project-specific role that matches
-            final SortedMap<Role, Set<String>> roles = rbas.getGrantedRoles(RoleBasedAuthorizationStrategy.PROJECT);
+            final SortedMap<Role, Set<String>> roles = rbas.getGrantedRoles(RoleType.Project);
             badList = new ArrayList<>(roles.size());
             if (StringUtils.isNotBlank(name)) {
                 for (SortedMap.Entry<Role, Set<String>> entry: roles.entrySet())  {
--- a/src/main/java/org/jenkinsci/plugins/rolestrategy/RoleBasedProjectNamingStrategy.java
+++ b/src/main/java/org/jenkinsci/plugins/rolestrategy/RoleBasedProjectNamingStrategy.java
@@ -1,5 +1,7 @@
 package org.jenkinsci.plugins.rolestrategy;
 
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
 import com.michelin.cio.hudson.plugins.rolestrategy.Messages;
 import com.michelin.cio.hudson.plugins.rolestrategy.Role;
 import com.michelin.cio.hudson.plugins.rolestrategy.RoleBasedAuthorizationStrategy;
@@ -9,13 +11,18 @@ import hudson.model.Item;
 import hudson.security.AuthorizationStrategy;
 import jenkins.model.Jenkins;
 import jenkins.model.ProjectNamingStrategy;
+import org.acegisecurity.Authentication;
+import org.acegisecurity.acls.sid.PrincipalSid;
+import org.acegisecurity.userdetails.UserDetails;
 import org.apache.commons.lang.StringUtils;
 import org.kohsuke.stapler.DataBoundConstructor;
 
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Set;
 import java.util.SortedMap;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 
 /**
@@ -28,6 +35,12 @@ public class RoleBasedProjectNamingStrategy extends ProjectNamingStrategy implem
 
     private final boolean forceExistingJobs;
 
+    private static final Cache<String, UserDetails> cache = CacheBuilder.newBuilder()
+            .softValues()
+            .maximumSize(Settings.USER_DETAILS_CACHE_MAX_SIZE)
+            .expireAfterWrite(Settings.USER_DETAILS_CACHE_EXPIRATION_TIME_SEC, TimeUnit.SECONDS)
+            .build();
+
     @DataBoundConstructor
     public RoleBasedProjectNamingStrategy(boolean forceExistingJobs) {
         this.forceExistingJobs = forceExistingJobs;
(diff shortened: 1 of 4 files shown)


nightowlengineer:
> An autotest would be great, but it looks good to me as is.
> Thanks!

Great - I'll try and add something tonight, otherwise it'll be a week or two before I can take a look at it again. Feel free to merge in the meantime if you like though.

nightowlengineer:
Well that was fun 😅 First time I've delved into any core or plugin code let alone the test side, so any comments/feedback on the way I've implemented the tests greatly appreciated.

nightowlengineer:
Re the failing CI:

I've fixed this [on a branch](https://github.com/nightowlengineer/role-strategy-plugin/compare/fix-rename-bug...nightowlengineer:fix-rename-bug-newer-lts) and am happy to merge into this if the bump in the Jenkins version is acceptable - it's only needed for the tests, not for users. It's due to changes to how renames are handled in >= 2.110 (see: [JENKINS-22936](https://issues.jenkins-ci.org/browse/JENKINS-22936)) which the CI is using.

src/test/java/org/jenkinsci/plugins/rolestrategy/util/TemporaryUserSession.java
```diff
@@ -0,0 +1,22 @@
+package org.jenkinsci.plugins.rolestrategy.util;
+
+import hudson.model.User;
+import org.acegisecurity.Authentication;
+import org.acegisecurity.context.SecurityContext;
+import org.acegisecurity.context.SecurityContextHolder;
+
+public class TemporaryUserSession implements AutoCloseable {
```
AbhyudayaSharma: Hi @nightowlengineer ! I think what you're trying to do already exists: [`ACL.as(User.get("username"))`](https://javadoc.jenkins.io/hudson/security/ACL.html#as-hudson.model.User-).

nightowlengineer: I figured it must have been done somewhere before but didn't look hard enough obviously! Will refactor to use that, cheers for the pointer!

nightowlengineer: Fixed.

nightowlengineer:
I can remove the mocks if the Jenkins core version is bumped - there's a doConfirmRename method after 2.110 which is just a simple string (see comment further up). Or can just test via the naming strategy as suggested above. I'd prefer the version bump (tech debt etc) - what would you prefer?

oleg-nenashev:
I am not against bumping the core. There were numerous security fixes after 2.121.x (LTS with the desired API), and I would expect major users of the plugin to have recent LTS versions.

@runzexia WDYT? If you prefer to keep the current LTS target, let's just keep mock tests for now (the failing test needs a fix tho)

nightowlengineer:
Have pushed up an additional commit with the core bumped and tests simplified as a result - happy to revert or squash after a decision is made.

src/test/java/org/jenkinsci/plugins/rolestrategy/RoleStrategyTest.java
```diff
@@ -3,36 +3,34 @@
 import com.cloudbees.hudson.plugins.folder.Folder;
 import com.michelin.cio.hudson.plugins.rolestrategy.Role;
 import com.michelin.cio.hudson.plugins.rolestrategy.RoleBasedAuthorizationStrategy;
-import hudson.model.Computer;
-import hudson.model.FreeStyleProject;
-import hudson.model.Item;
-import hudson.model.User;
+import hudson.model.*;
```
runzexia: Should not import *

src/test/java/org/jenkinsci/plugins/rolestrategy/RoleStrategyTest.java
```diff
@@ -3,36 +3,34 @@
...
 import org.junit.Test;
 import org.jvnet.hudson.test.Issue;
 
-import static io.jenkins.plugins.casc.misc.Util.getJenkinsRoot;
-import static io.jenkins.plugins.casc.misc.Util.toStringFromYamlFile;
-import static io.jenkins.plugins.casc.misc.Util.toYamlString;
+import java.util.Map;
+import java.util.Set;
+
+import static io.jenkins.plugins.casc.misc.Util.*;
```
runzexia: also

src/test/java/org/jenkinsci/plugins/rolestrategy/RoleStrategyTest.java
```diff
@@ -3,36 +3,34 @@
...
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.instanceOf;
 import static org.hamcrest.core.Is.is;
 import static org.jenkinsci.plugins.rolestrategy.PermissionAssert.assertHasNoPermission;
 import static org.jenkinsci.plugins.rolestrategy.PermissionAssert.assertHasPermission;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertThat;
+import static org.junit.Assert.*;
```
runzexia: also

src/main/java/org/jenkinsci/plugins/rolestrategy/RoleBasedProjectNamingStrategy.java
```diff
@@ -37,24 +50,36 @@ public RoleBasedProjectNamingStrategy(boolean forceExistingJobs) {
...
+            final SortedMap<Role, Set<String>> gRole = rbas.getGrantedRoles(RoleBasedAuthorizationStrategy.GLOBAL);
+            for (final SortedMap.Entry<Role, Set<String>> entry : gRole.entrySet()) {
+                if (entry.getKey().hasPermission(Item.CREATE)) {
+                    globalRolePermitted = true;
+                    break;
+                }
+            }
+
+            if (!globalRolePermitted) {
+                throw new Failure(Messages.RoleBasedProjectNamingStrategy_NoPermissions());
```
runzexia: I am not sure here, when searching for the global role does not have permission, will it also search for the project role?

runzexia: In my mind, users with global creation roles can create items of any name.
@oleg-nenashev

nightowlengineer: This PR was spawned from the code in #16 and I didn't want to change too much. However I agree the logic does feel odd at the moment and differs from the current released code.

I guess it depends on whether having a global role is intended to completely override any other project roles that might exist/match for a user?

runzexia: I think the global role is to cover the project role.
https://github.com/jenkinsci/role-strategy-plugin/blob/master/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java#L131

src/main/java/org/jenkinsci/plugins/rolestrategy/RoleBasedProjectNamingStrategy.java
```diff
@@ -37,24 +50,36 @@ public RoleBasedProjectNamingStrategy(boolean forceExistingJobs) {
...
-            //firstly check global role
-            SortedMap<Role, Set<String>> gRole = rbas.getGrantedRoles(RoleBasedAuthorizationStrategy.GLOBAL);
-            for (SortedMap.Entry<Role, Set<String>> entry: gRole.entrySet()){
-                if (entry.getKey().hasPermission(Item.CREATE))
-                    return;
+        final AuthorizationStrategy auth = Jenkins.getInstance().getAuthorizationStrategy();
+        if (auth instanceof RoleBasedAuthorizationStrategy) {
+            final RoleBasedAuthorizationStrategy rbas = (RoleBasedAuthorizationStrategy) auth;
+
+            // Check whether the user has a global role allowing them to carry out this action
```
oleg-nenashev:
```suggestion
            // Check whether the user has a global role allowing them to carry out this action
            // TODO(oleg-nenashev): This is an odd hack which should be fixed by improving Jenkins core APIs
            // This code comes from the fact that the users will not be even able to see the Jenkins Web UI or to invoke the create action if they have no this permission. 
            // In the current Jenkins permission model a user always needs global permission to check roles before project permissions get applied
```

src/main/java/org/jenkinsci/plugins/rolestrategy/RoleBasedProjectNamingStrategy.java
oleg-nenashev: Suggested the change above.
Ideally we should create a ticket on the Jenkins core side to improve the permission model. But so far a global permission is required if a user wants to see the create dialog.

nightowlengineer:
OK I think I've resolved all the review comments and have squashed the commits back down - let me know if there's anything else to sort out 😄

src/main/java/org/jenkinsci/plugins/rolestrategy/RoleBasedProjectNamingStrategy.java
runzexia: In folder based management, users can have permission to create projects under the folder without having to create global permissions.
Not sure if this will affect some users, the best is to create a ticket in Jenkins Core

nightowlengineer: I'll try and give that scenario a shot later today (this PR + folders plugin) to see exactly what happens.

nightowlengineer: So umm, I couldn't get folder naming conventions to work on master... I'm sure I've done it before, but I can't see how it works because you only get the item name to check (like 'job') not the full path (like 'folder/job'). I'm obviously missing something though 🤔

runzexia: try with folder "a" and role: 
project role1: "^a$"
project role2: "^a/.*"
both two roles have create item permission

nightowlengineer: Fairly sure I tried that but will give it another shot - won't be until next week unfortunately.

nightowlengineer: Just to confirm I'm still planning on picking this up again this week.

mawinter69:
Fixed via #179

PR #77, opened 2019-06-13
Add new benchmark for RoleMap#newMatchingRoleMap

AbhyudayaSharma:
Benchmarks `newMatchingRoleMap` as all of the regex matching takes place there
Also use `Sets.newHashSet` instead of `new HashSet(Arrays.asList(...))`.

profile2


oleg-nenashev:
I would call it a dedicated test for Folder-based permission use-case

PR #81, opened 2019-06-20
Performance: Cache RoleMaps produced by `RoleMap#newMatchingRoleMap()`

AbhyudayaSharma:
From our discussion yesterday on the Gitter chat and since the newMatchingRoleMap is called for every ACL request and a lot of permissions are checked for a project when the home page loads, having a cache avoids checking regular expressions again for every time permission checks are called for a given Item.

### DIFF
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
@@ -203,8 +203,8 @@ public class RoleMap {
   public void addRole(Role role) {
       if (this.getRole(role.getName()) == null) {
           this.grantedRoles.put(role, new CopyOnWriteArraySet<>());
+          matchingRoleMapCache.invalidateAll();
       }
-
   }
 
   /**
@@ -225,11 +225,10 @@ public class RoleMap {
    * @since 2.6.0
    */
   public void unAssignRole(Role role, String sid) {
-    if (this.hasRole(role)) {
-      if (this.grantedRoles.get(role).contains(sid)) {
-        this.grantedRoles.get(role).remove(sid);
+      Set<String> sids = grantedRoles.get(role);
+      if (sids != null) {
+        sids.remove(sid);
       }
-    }
   }
 
   /**
@@ -304,6 +303,7 @@ public class RoleMap {
    */
   public void removeRole(Role role){
       this.grantedRoles.remove(role);
+      matchingRoleMapCache.invalidateAll();
   }
   
 
@@ -364,13 +364,14 @@ public class RoleMap {
   }
 
   /**
-   * Create a sub-map of the current {@link RoleMap} containing only the
-   * {@link Role}s matching the given pattern.
-   * @param namePattern The pattern to match
-   * @return A {@link RoleMap} containing only {@link Role}s matching the given name
+   * Create a sub-map of this {@link RoleMap} containing {@link Role}s that are applicable
+   * on the given {@code itemName}.
+   *
+   * @param itemName the name of the {@link hudson.model.AbstractItem} or {@link hudson.model.Computer}
+   * @return A {@link RoleMap} containing roles that are applicable on the itemName
    */
-  public RoleMap newMatchingRoleMap(String namePattern) {
-    RoleMap cachedRoleMap = matchingRoleMapCache.getIfPresent(namePattern);
+  public RoleMap newMatchingRoleMap(String itemName) {
+    RoleMap cachedRoleMap = matchingRoleMapCache.getIfPresent(itemName);
     if (cachedRoleMap != null) {
       return cachedRoleMap;
     }
@@ -378,42 +379,16 @@ public class RoleMap {
     SortedMap<Role, Set<String>> roleMap = new TreeMap<>();
     new RoleWalker() {
       public void perform(Role current) {
-        Matcher m = current.getPattern().matcher(namePattern);
+        Matcher m = current.getPattern().matcher(itemName);
         if (m.matches()) {
           roleMap.put(current, grantedRoles.get(current));
         }
       }
     };
+
     RoleMap newMatchingRoleMap = new RoleMap(roleMap);
-    matchingRoleMapCache.put(namePattern, newMatchingRoleMap);
+    matchingRoleMapCache.put(itemName, newMatchingRoleMap);
     return newMatchingRoleMap;
-  }
-
-  /**
-   * Get all the roles holding the given permission.
-   * @param permission The permission you want to check
-   * @return A Set of Roles holding the given permission
-   */
-  private Set<Role> getRolesHavingPermission(final Permission permission) {
-    final Set<Role> roles = new HashSet<>();
-    final Set<Permission> permissions = new HashSet<>();
-    Permission p = permission;
-
-    // Get the implying permissions
-    for (; p!=null; p=p.impliedBy) {
-      permissions.add(p);
-    }
-    // Walk through the roles, and only add the roles having the given permission,
-    // or a permission implying the given permission
-    new RoleWalker() {
-      public void perform(Role current) {
-        if (current.hasAnyPermission(permissions)) {
-          roles.add(current);
-        }
-      }
-    };
-
-    return roles;
   }
 
   /**
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
@@ -86,6 +86,16 @@ public class RoleMap {
           .expireAfterWrite(Settings.USER_DETAILS_CACHE_EXPIRATION_TIME_SEC, TimeUnit.SECONDS)
           .build();
 
+  /**
+   * {@link RoleMap}s are created again and again using {@link RoleMap#newMatchingRoleMap(String)}
+   * for different permissions for the same {@code itemName}, so cache them and avoid wasting time
+   * matching regular expressions.
+   */
+  private final Cache<String, RoleMap> matchingRoleMapCache = CacheBuilder.newBuilder()
+          .softValues()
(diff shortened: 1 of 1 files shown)


src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -86,6 +86,16 @@
           .expireAfterWrite(Settings.USER_DETAILS_CACHE_EXPIRATION_TIME_SEC, TimeUnit.SECONDS)
           .build();
 
+  /**
+   * {@link RoleMap}s are created again and again using {@link RoleMap#newMatchingRoleMap(String)}
+   * for different permissions for the same {@code itemName}, so cache them and avoid wasting time
+   * matching regular expressions.
+   */
+  private final Cache<String, RoleMap> matchingRoleMapCache = CacheBuilder.newBuilder()
```
AbhyudayaSharma: Do we continue with the Cache from Guava (avoid adding any dependencies) or use something else like [JCS](http://commons.apache.org/proper/commons-jcs/), [Ehcache](https://github.com/ehcache/ehcache3) or something else?

AbhyudayaSharma:
Results compared to the benchmarks carried out in #80 : 

![image](https://user-images.githubusercontent.com/11471599/59837236-8a712c80-936a-11e9-84a6-29cc8558714f.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -364,58 +364,33 @@ public void removeRole(Role role){
...
-   * @return A {@link RoleMap} containing only {@link Role}s matching the given name
+   * Create a sub-map of this {@link RoleMap} containing {@link Role}s that are applicable
+   * on the given {@code itemName}.
+   *
+   * @param itemName the name of the {@link hudson.model.AbstractItem} or {@link hudson.model.Computer}
+   * @return A {@link RoleMap} containing roles that are applicable on the itemName
    */
-  public RoleMap newMatchingRoleMap(String namePattern) {
-    RoleMap cachedRoleMap = matchingRoleMapCache.getIfPresent(namePattern);
+  public RoleMap newMatchingRoleMap(String itemName) {
```
AbhyudayaSharma: I've changed the JavaDoc because I don't think it correctly told what was going on.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -364,58 +364,33 @@ public void removeRole(Role role){
...
+    matchingRoleMapCache.put(itemName, newMatchingRoleMap);
     return newMatchingRoleMap;
   }
 
-  /**
-   * Get all the roles holding the given permission.
-   * @param permission The permission you want to check
-   * @return A Set of Roles holding the given permission
-   */
-  private Set<Role> getRolesHavingPermission(final Permission permission) {
```
AbhyudayaSharma: I've removed it because this method was not being used anywhere.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -203,8 +203,8 @@ public SidACL getACL(RoleType roleType, AccessControlled controlledItem) {
   public void addRole(Role role) {
       if (this.getRole(role.getName()) == null) {
           this.grantedRoles.put(role, new CopyOnWriteArraySet<>());
+          matchingRoleMapCache.invalidateAll();
```
AbhyudayaSharma: This cache would only be affected when a change to `grantedRoles` is made because the `RoleMap` returned contains only those roles applicable for a given name. So it won't be necessary to use `ItemListener` in this case.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
AbhyudayaSharma: It was refactored out in #54 and has no references to it now.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
oleg-nenashev: I agree that the original Javadoc was wrong, 
Maybe we could rename it to `itemNamePrefix` or so, because actually it does some kind of "pattern matching", just not the regexp/wildcard/whatever. It just takes a string and verifies it against regexp, but it does not really have to be a full item name. For better or worse...

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
oleg-nenashev: It is fine to use Guava as long as we do not needed newer versions.
Maybe we could have a research task for comparing cache implementations with benchmarks in later phases

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -86,6 +86,16 @@
...
 
+  /**
+   * {@link RoleMap}s are created again and again using {@link RoleMap#newMatchingRoleMap(String)}
+   * for different permissions for the same {@code itemName}, so cache them and avoid wasting time
+   * matching regular expressions.
+   */
+  private final Cache<String, RoleMap> matchingRoleMapCache = CacheBuilder.newBuilder()
+          .softValues()
+          .maximumSize(Settings.USER_DETAILS_CACHE_MAX_SIZE)
+          .expireAfterWrite(Settings.USER_DETAILS_CACHE_EXPIRATION_TIME_SEC, TimeUnit.SECONDS)
```
oleg-nenashev: Maybe bump to 1 hour here? We do not expect roles to change often

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
oleg-nenashev: agreed

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
oleg-nenashev: It looks like a good API method, but we can always recover it later if somebody needs it

oleg-nenashev:
Test failure is really odd here, and last reopen did not trigger the build. Let me try

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -86,6 +86,16 @@
...
           .build();
 
+  /**
+   * {@link RoleMap}s are created again and again using {@link RoleMap#newMatchingRoleMap(String)}
+   * for different permissions for the same {@code itemNamePrefix}, so cache them and avoid wasting time
+   * matching regular expressions.
+   */
+  private final Cache<String, RoleMap> matchingRoleMapCache = CacheBuilder.newBuilder()
+          .softValues()
+          .maximumSize(Settings.USER_DETAILS_CACHE_MAX_SIZE)
```
AbhyudayaSharma: The default value for the max size of the cache is 100. If there are more than 100 objects managed by Role Strategy, the cache would start evicting entries. Also, the max size of the cache cannot be increased after its creation. Should we care about this situation and increase the size of the cache using an `ItemListener` and a `ComputerListener` at the cost of higher memory usage?

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -86,6 +86,16 @@
           .expireAfterWrite(Settings.USER_DETAILS_CACHE_EXPIRATION_TIME_SEC, TimeUnit.SECONDS)
           .build();
 
+  /**
+   * {@link RoleMap}s are created again and again using {@link RoleMap#newMatchingRoleMap(String)}
+   * for different permissions for the same {@code itemNamePrefix}, so cache them and avoid wasting time
+   * matching regular expressions.
+   */
+  private final Cache<String, RoleMap> matchingRoleMapCache = CacheBuilder.newBuilder()
```
AbhyudayaSharma: The matching RoleMaps that are created also have a cache in them which is not needed. Do we reorganize the class structure to avoid it or don't care about a little more memory use?

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
oleg-nenashev: +1. The objects are not big, so bumping to 2048 or so would make sense

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
oleg-nenashev: I would submit a follow-up ticket for it. Memory usage is not a problem for the plugin ATM

scrwr:
Hi. This has some really odd implication as it seems. A typical workflow is:
1. create a new role
2. assign the role 

But role assignment doesn't invalidate the cache and hence, the user sticks with his old roleset for an hour.

odilontalk:
Yeah, facing the same issue mentioned by @scrwr [here](https://github.com/jenkinsci/role-strategy-plugin/pull/81#issuecomment-579212814). 

There is any command that i can run to invalidate cache? Maybe in Jenkins Script Console?

oleg-nenashev:
I plan to cut a releas soon, sorry for delays

grickit:
@odilontalk I just ran into this myself and confirmed (as implied by @scrwr's comment) that creating a new arbitrary dummy role and hitting "apply" will clear/update the cached permissions. You can then delete that role.

PR #83, opened 2019-06-24
[JENKINS-58399] Performance: Avoid finding implying permissions on each `hasPermission()` call

AbhyudayaSharma:
The permissions that a imply a permission are calculated every time. This pull request caches the results in a  `ConcurrentHashMap`.

### DIFF
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
@@ -30,20 +30,22 @@ import com.synopsys.arc.jenkins.plugins.rolestrategy.Macro;
 import com.synopsys.arc.jenkins.plugins.rolestrategy.RoleMacroExtension;
 import com.synopsys.arc.jenkins.plugins.rolestrategy.RoleType;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import hudson.model.Item;
 import hudson.model.Items;
 import hudson.model.User;
 import hudson.security.AccessControlled;
 import hudson.security.Permission;
 import hudson.security.SidACL;
-import hudson.model.Item;
 import jenkins.model.Jenkins;
-
 import org.acegisecurity.BadCredentialsException;
 import org.acegisecurity.GrantedAuthority;
 import org.acegisecurity.acls.sid.Sid;
 import org.acegisecurity.userdetails.UserDetails;
 import org.jenkinsci.plugins.rolestrategy.Settings;
+import org.jenkinsci.plugins.rolestrategy.permissions.DangerousPermissionHandlingMode;
 import org.jenkinsci.plugins.rolestrategy.permissions.PermissionHelper;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.springframework.dao.DataAccessException;
 
@@ -107,6 +109,14 @@ public class RoleMap {
     public RoleMap(@Nonnull SortedMap<Role,Set<String>> grantedRoles) {
         this.grantedRoles = new ConcurrentSkipListMap<Role, Set<String>>(grantedRoles);
     }
+
+  /**
+   * Invalidate the dangerous permissions from the cache when the {@link DangerousPermissionHandlingMode} changes.
+   */
+  @Restricted(NoExternalUse.class)
+  public static void invalidateDangerousPermissions() {
+    PermissionHelper.DANGEROUS_PERMISSIONS.forEach(implyingPermissionCache::remove);
+  }
 
   /**
    * Check if the given sid has the provided {@link Permission}.
@@ -188,15 +198,17 @@ public class RoleMap {
    * @return a set of permissions that imply this permission (including itself)
    */
   private static Set<Permission> cacheImplyingPermissions(Permission permission) {
+    Set<Permission> implyingPermissions;
     if (PermissionHelper.isDangerous(permission)) {
       /* if this is a dangerous permission, fall back to Administer unless we're in compat mode */
-      permission = Jenkins.ADMINISTER;
-    }
+      implyingPermissions = getImplyingPermissions(Jenkins.ADMINISTER);
+    } else {
+      implyingPermissions = new HashSet<>();
 
-    Set<Permission> implyingPermissions = new HashSet<>();
-    // Get the implying permissions
-    for (Permission p = permission; p != null; p = p.impliedBy) {
-      implyingPermissions.add(p);
+      // Get the implying permissions
+      for (Permission p = permission; p != null; p = p.impliedBy) {
+        implyingPermissions.add(p);
+      }
     }
 
     implyingPermissionCache.put(permission, implyingPermissions);
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
@@ -60,6 +60,8 @@ import java.util.SortedMap;
 import java.util.SortedSet;
 import java.util.TreeMap;
 import java.util.TreeSet;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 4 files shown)


src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -79,6 +81,12 @@
   private final SortedMap <Role,Set<String>> grantedRoles;
 
   private static final Logger LOGGER = Logger.getLogger(RoleMap.class.getName());
+
+  private static final ConcurrentMap<Permission, Set<Permission>> implyingPermissionCache = new ConcurrentHashMap<>();
```
AbhyudayaSharma: Using a `ConcurrentHashMap` because I could not find any way to remove a `Permission` so there should not be a need to 'expire' the cache after a given time.

AbhyudayaSharma:
Results, compared with #80 again:

image

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
oleg-nenashev: In Jenkins we usually use uppercase names for static final variables, but this is not how https://web.archive.org/web/20120911192801/developers.sun.com/sunstudio/products/archive/whitepapers/java-style.pdf written. So fine with me

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -79,6 +81,12 @@
...
+
+  static {
+    Permission.getAll().forEach(RoleMap::cacheImplyingPermissions);
```
oleg-nenashev: What if the permissions change during the execution, e.g. by loading a plugin with new permissions?

AbhyudayaSharma: If the permission is not in the cache, it calculates the implying permissions and stores them back in the map:

https://github.com/jenkinsci/role-strategy-plugin/blob/b84b0ec30e19719f235ae333d566263bf182e212/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java#L175-L182

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -127,29 +127,27 @@ public void perform(Role current) {
...
+              hasPermission[0] = true;
               abort();
-              return ;
             }
           } else if (Settings.TREAT_USER_AUTHORITIES_AS_ROLES) {
             try {
               UserDetails userDetails = cache.getIfPresent(sid);
               if (userDetails == null) {
-                userDetails = Jenkins.getActiveInstance().getSecurityRealm().loadUserByUsername(sid);
+                userDetails = Jenkins.getInstance().getSecurityRealm().loadUserByUsername(sid);
```
res0nance: Probably want to use Jenkins.get() here since getInstance() is deprecated and can return null

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -104,18 +112,10 @@ public RoleMap(@Nonnull SortedMap<Role,Set<String>> grantedRoles) {
...
-    final Set<Permission> permissions = new HashSet<>();
-    final Permission per = p;
-    final boolean[] temp = {false};
-    // Get the implying permissions
-    for (; p!=null; p=p.impliedBy) {
-      permissions.add(p);
-    }
+  private boolean hasPermission(String sid, Permission permission, RoleType roleType, AccessControlled controlledItem) {
+    final Set<Permission> permissions = getImplyingPermissions(permission);
+    final boolean[] hasPermission = { false };
```
res0nance: Wondering why hasPermission is an array when only 1 boolean is ever used

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
AbhyudayaSharma: Role Strategy uses an old version of Jenkins, so `Jenkins.get()` is not available to use right now.

casz: Not the case for the current jenkins.version they use: 
https://github.com/jenkinsci/role-strategy-plugin/blob/master/pom.xml#L44

Though perhaps a good time to upgrade

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
AbhyudayaSharma: I guess that's just to make the reference final but still mutable. Seems to come directly from https://stackoverflow.com/a/4520163/6306974

src/main/java/org/jenkinsci/plugins/rolestrategy/permissions/DangerousPermissionHandlingMode.java
```diff
@@ -82,4 +81,15 @@
...
         return CURRENT;
     }
+
+    /**
+     * Set the {@link DangerousPermissionHandlingMode}
+     *
+     * @param mode the {@code DangerousPermissionHandlingMode} to be set
+     */
+    @Restricted(NoExternalUse.class)
+    public static void setCURRENT(@Nonnull DangerousPermissionHandlingMode mode) {
```
AbhyudayaSharma: Had to use uppercase `CURRENT` to allow backward compatibility from the following Groovy script:
```groovy
import org.jenkinsci.plugins.rolestrategy.permissions.*

DangerousPermissionHandlingMode.CURRENT = DangerousPermissionHandlingMode.UNDEFINED
```

src/main/java/org/jenkinsci/plugins/rolestrategy/permissions/DangerousPermissionHandlingMode.java
```diff
@@ -61,9 +61,8 @@
     public static final String PROPERTY_NAME = DangerousPermissionHandlingMode.class.getName() + ".enableDangerousPermissions";
     
     @Nonnull
-    @Restricted(NoExternalUse.class)
     @SuppressFBWarnings(value = "MS_SHOULD_BE_REFACTORED_TO_BE_FINAL", justification = "Groovy script console access")
-    public static /* allow script access */ DangerousPermissionHandlingMode CURRENT;
+    private static DangerousPermissionHandlingMode CURRENT;
```
AbhyudayaSharma: Made it private. Now the access is through getters and setters allowing us to have hooks when this is changed and invalidate dangerous permissions from the cache.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
runzexia: Not an important issue, but since he won't expire, we might change a name like "allImplyingPermission"

AbhyudayaSharma: @runzexia It may not contain **all** implying permissions at a given time. Say when a plugin adds new permissions during run time (as mentioned by Oleg below), the permissions may not be present in this cache until a request for that permission has been made. I hope that makes some sense.

AbhyudayaSharma:
Wow, now #81 has been merged, the results are looking much better:

image

casz:
Wow that is some seriously awesome benchmark improvements

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
oleg-nenashev: 2.138.x would be fidne with me, and it supports the required API

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -79,6 +83,12 @@
   private final SortedMap <Role,Set<String>> grantedRoles;
 
   private static final Logger LOGGER = Logger.getLogger(RoleMap.class.getName());
+
+  private static final ConcurrentMap<Permission, Set<Permission>> implyingPermissionCache = new ConcurrentHashMap<>();
```
oleg-nenashev: It causes interlocks between RoleMap instances, and this cache may also roll over in tests since it is a static field (Maven Surefire plugin runs tests in the same Java context). Is it an intended behavior?

oleg-nenashev:
https://github.com/jenkinsci/extended-read-permission-plugin/blob/master/src/main/java/hudson/plugins/extendedread/PluginImpl.java#L32 for example of permissions management in runtime. If a system groovy script disables permissions in such way, caching will not reflect it.

w.r.t. caching. It can be either opt-in or opt-out. CC @jenkinsci/gsoc2019-role-strategy

oleg-nenashev:
Also notes for reviewers. https://javadoc.jenkins.io/hudson/security/Permission.html is final, and hence permission implementations cannot define the `impliedBy` behavior in runtime unless they use reflection. So the static caching should be fine unless we use Groovy magic or whatever.

CC Jenkins Security team, because they might be able to comment on the risks we might be missing. @daniel-beck @Wadeck @jeffret-b @jvz

oleg-nenashev:
My plan is to integrate it on Thursday if no negative feedback

oleg-nenashev:
Thanks @AbhyudayaSharma !

PR #85, opened 2019-06-26
Refactor RoleBasedAuthorizationStrategy to use Enum keys

AbhyudayaSharma:
`grantedRoles` in `RoleBasedAuthorizationStrategy` used String keys for getting values from the map. This changes it to use `RoleType` enum keys.

### DIFF
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
@@ -62,7 +62,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.SortedMap;
-import java.util.logging.Logger;
 import java.util.regex.Pattern;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
@@ -92,18 +91,36 @@ public class RoleBasedAuthorizationStrategy extends AuthorizationStrategy {
   public final static String SLAVE     = "slaveRoles";
   public final static String MACRO_ROLE = "roleMacros";
   public final static String MACRO_USER  = "userMacros";
-  
-  private static final Logger LOGGER = Logger.getLogger(RoleBasedAuthorizationStrategy.class.getName());
-  
-  /** {@link RoleMap}s associated to each {@link AccessControlled} class */
-  private final Map <String, RoleMap> grantedRoles;
+
+  /**
+   * {@link RoleMap}s for each {@link RoleType}
+   */
+  private final Map<RoleType, RoleMap> roleMaps = new CopyOnWriteMap.Hash<>();
 
   public RoleBasedAuthorizationStrategy() {
-      this.grantedRoles = new CopyOnWriteMap.Hash<>();
+    Arrays.stream(RoleType.values()).forEach(roleType -> roleMaps.put(roleType, new RoleMap()));
   }
 
+  /**
+   * Creates a new {@link RoleBasedAuthorizationStrategy}
+   *
+   * @param grantedRoles the roles in the strategy
+   * @deprecated Use {@link RoleBasedAuthorizationStrategy#RoleBasedAuthorizationStrategy(HashMap)}
+   */
+  @Deprecated
   public RoleBasedAuthorizationStrategy(Map<String, RoleMap> grantedRoles) {
-      this.grantedRoles = new CopyOnWriteMap.Hash<>(grantedRoles);
+    this();
+    grantedRoles.forEach((type, roleMap) -> roleMaps.put(RoleType.fromString(type), roleMap));
+  }
+
+  /**
+   * Creates a new {@link RoleBasedAuthorizationStrategy}
+   *
+   * @param roleMaps a subset of the the role
+   */
+  public RoleBasedAuthorizationStrategy(HashMap<RoleType, RoleMap> roleMaps) {
+    this();
+    roleMaps.forEach(this.roleMaps::put);
   }
 
     /**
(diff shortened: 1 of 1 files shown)


src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -92,49 +91,61 @@
...
-  
-  private static final Logger LOGGER = Logger.getLogger(RoleBasedAuthorizationStrategy.class.getName());
-  
-  /** {@link RoleMap}s associated to each {@link AccessControlled} class */
-  private final Map <String, RoleMap> grantedRoles;
+
+  /**
+   * {@link RoleMap}s for each {@link RoleType}
+   */
+  private final Map<RoleType, RoleMap> roleMaps = new CopyOnWriteMap.Hash<>();
```
AbhyudayaSharma: Renamed it to `roleMaps` because `grantedRoles` did not seem right for this situation.

Using an enum might not make much of a difference in hashing performance because `String` class caches it's hash code too.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -369,7 +355,7 @@ public void doRemoveRoles(@QueryParameter(required = true) String type,
                               @QueryParameter(required = true) String roleNames) throws IOException {
```
AbhyudayaSharma: Can we use the `RoleType` enum as a query parameter instead of parsing the string ourselves?

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -575,8 +562,7 @@ public void marshal(Object source, HierarchicalStreamWriter writer, MarshallingC
 
       public Object unmarshal(HierarchicalStreamReader reader, final UnmarshallingContext context) {
         final RoleBasedAuthorizationStrategy strategy = create();
-        boolean showDangerousPermissionsDefined = false;
```
AbhyudayaSharma: This was unused.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
casz: I believe stapler will do the conversion

AbhyudayaSharma:
image

The decline is <= 5% in all cases, so there is no actual performance difference from this pull request.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -149,25 +173,25 @@ public ACL getACL(Job<?,?> project) {
 
     @Override
     public ACL getACL(AbstractItem project) {
-      return getACL(PROJECT, project.getFullName(), RoleType.Project, project);
+      return getACL(project.getFullName(), RoleType.Project, project);
```
oleg-nenashev: Here (and in all type-specific methods) we could avoid calling the generic method and access fields directly so that we get some performance advantage

AbhyudayaSharma: Done! Saves us from unnecessarily calling the `getRoleMap()` method once.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -194,56 +216,29 @@ public ACL getACL(Computer computer) {
    */
   @CheckForNull
   public Set<String> getSIDs(String type) {
```
oleg-nenashev: Add new API as well? Can be done in a follow-up

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -92,82 +90,99 @@
...
-   /**
-   * Get the specific ACL for projects.
-   * @param project The access-controlled project
-   * @return The project specific ACL
+   * Get the {@link RoleMap} corresponding to the {@link RoleType}
+   *
+   * @param roleType the type of the role
+   * @return the {@link RoleMap} corresponding to the {@code roleType}
    */
+  private RoleMap getRoleMap(RoleType roleType) {
```
oleg-nenashev: It might make sense to start annotating methods just in case. E.g. this one returns `@Nonnull`. Also, it makes sense to put the exception to Javadoc

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -92,82 +90,99 @@
...
   public final static String MACRO_USER  = "userMacros";
-  
-  private static final Logger LOGGER = Logger.getLogger(RoleBasedAuthorizationStrategy.class.getName());
-  
-  /** {@link RoleMap}s associated to each {@link AccessControlled} class */
-  private final Map <String, RoleMap> grantedRoles;
+
+  private final RoleMap slaveRoles;
+  private final RoleMap globalRoles;
+  private final RoleMap projectRoles;
```
oleg-nenashev: itemRoles ?

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -92,82 +90,99 @@
   public final static String SLAVE     = "slaveRoles";
   public final static String MACRO_ROLE = "roleMacros";
   public final static String MACRO_USER  = "userMacros";
-  
-  private static final Logger LOGGER = Logger.getLogger(RoleBasedAuthorizationStrategy.class.getName());
-  
-  /** {@link RoleMap}s associated to each {@link AccessControlled} class */
-  private final Map <String, RoleMap> grantedRoles;
+
+  private final RoleMap slaveRoles;
```
oleg-nenashev: agentRoles or nodeRoles ?

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
AbhyudayaSharma: Outdated: this is has been completely removed.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
AbhyudayaSharma: Didn't work because the enum names were not the same as the strings. Thanks anyway!

AbhyudayaSharma:
> I wonder what is the performance difference now, but I hope it has improved a bit

No significant improvements; almost the same as the image above.

PR #91, opened 2019-07-11
[JENKINS-58227] Ignore dangerous permissions on Configuration Export

AbhyudayaSharma:
When the dangerous permission handling mode is disabled, we ignore dangerous permissions because during the permission check, the permissions are deferred to `Jenkins.ADMINISTER`.

### DIFF
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/Role.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/Role.java
@@ -95,7 +95,7 @@ public final class Role implements Comparable {
   public Role(@Nonnull String name, @CheckForNull String pattern, @CheckForNull Set <String> permissionIds, @CheckForNull String description) {
       this(name,
            Pattern.compile(pattern != null ? pattern : GLOBAL_ROLE_PATTERN),
-           PermissionHelper.fromStrings(permissionIds),
+           PermissionHelper.fromStrings(permissionIds, true),
            description);
   }
 
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/Role.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/Role.java
@@ -32,6 +32,7 @@ import java.util.regex.Pattern;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.apache.commons.collections.CollectionUtils;
+import org.jenkinsci.plugins.rolestrategy.permissions.DangerousPermissionHandlingMode;
 import org.jenkinsci.plugins.rolestrategy.permissions.PermissionHelper;
 import org.kohsuke.stapler.DataBoundConstructor;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/rolestrategy/permissions/PermissionHelper.java
```diff
@@ -71,6 +74,21 @@ private PermissionHelper() {
      */
     @Nonnull
     public static Set<Permission> fromStrings(@CheckForNull Collection<String> permissionIds) throws SecurityException {
```
oleg-nenashev: Can be just removed - the `PermissionHelper ` is restricted

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/Role.java
```diff
@@ -95,7 +95,7 @@
   public Role(@Nonnull String name, @CheckForNull String pattern, @CheckForNull Set <String> permissionIds, @CheckForNull String description) {
       this(name,
            Pattern.compile(pattern != null ? pattern : GLOBAL_ROLE_PATTERN),
-           PermissionHelper.fromStrings(permissionIds),
+           PermissionHelper.fromStrings(permissionIds, true),
```
oleg-nenashev: I do not think it is a right fix, because it would impact the XML loading from what I see

AbhyudayaSharma: > My suggestion would be to just ignore dangerous permissions in the export when `DangerousPermissionHandlingMode.ENABLED` is false.

If the mode is `ENABLED`, none of the permissions are dangerous: https://github.com/jenkinsci/role-strategy-plugin/blob/f410ae394838137ed2f5a57020af204ca3aba7b3/src/main/java/org/jenkinsci/plugins/rolestrategy/permissions/PermissionHelper.java#L100-L107

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/Role.java
```diff
@@ -95,7 +96,7 @@
   public Role(@Nonnull String name, @CheckForNull String pattern, @CheckForNull Set <String> permissionIds, @CheckForNull String description) {
       this(name,
            Pattern.compile(pattern != null ? pattern : GLOBAL_ROLE_PATTERN),
-           PermissionHelper.fromStrings(permissionIds),
+           PermissionHelper.fromStrings(permissionIds, DangerousPermissionHandlingMode.getCurrent() != DangerousPermissionHandlingMode.ENABLED),
```
AbhyudayaSharma: Is this better?

oleg-nenashev: It does not look to solve the original issue: Dangerous permissions are ignored here when the mode is not enabled. I need to do some testing, but AFAICT it allows passing dangerous permissions, e.g. when loading the configuration from the disk

AbhyudayaSharma: Yes, with this fix the dangerous permissions will be allowed when loading from the disk (and during export), but if I recall correctly the same behavior happens when loading from `config.xml`. If that is not acceptable, I would agree with `WONTFIX`, like you had said during Friday's meeting.

nguyenhugh:
does anyone have an ETA on the fix? Purry Pleeze

AbhyudayaSharma:
> does anyone have an ETA on the fix? Purry Pleeze

@nguyenhugh In the meanwhile, you can enable dangerous permissions when you export

darxriggs:
@AbhyudayaSharma this pull request requires conflicts to be resolved.

Geo87:
Any chance can be  resolved and merged?  @AbhyudayaSharma

oleg-nenashev:
#131 fixes it by removing support for Dangerous permissions entirely

PR #94, opened 2019-07-18
[JENKINS-58536] Remove dependency on deprecated jquery-ui plugin

rdpa:
Remove deprecated jquery-ui dependency. Starting here to figure out what needs to be done next.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -87,9 +87,9 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.jenkins-ci.plugins</groupId>
-            <artifactId>jquery</artifactId>
-            <version>1.11.2-0</version>
+            <groupId>org.jenkins-ci.ui</groupId>
+            <artifactId>jquery-detached</artifactId>
+            <version>1.2.1</version>
         </dependency>
         <dependency>
             <groupId>io.jenkins</groupId>
@@ -106,5 +106,3 @@
         </dependency>
     </dependencies>
 </project>
-
-
--- a/pom.xml
+++ b/pom.xml
@@ -92,11 +92,6 @@
             <version>1.11.2-0</version>
         </dependency>
         <dependency>
-            <groupId>org.jenkins-ci.plugins</groupId>
-            <artifactId>jquery-ui</artifactId>
-            <version>1.0.1</version>
-        </dependency>
-        <dependency>
             <groupId>io.jenkins</groupId>
             <artifactId>configuration-as-code</artifactId>
             <version>${configuration-as-code.version}</version>
--- a/pom.xml
+++ b/pom.xml
@@ -87,11 +87,6 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.jenkins-ci.ui</groupId>
-            <artifactId>jquery-detached</artifactId>
-            <version>1.2.1</version>
-        </dependency>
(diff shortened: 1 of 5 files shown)


rdpa:
@oleg-nenashev Thanks for the quick response! If you would like I can take a look at that.

oleg-nenashev:
Fine with me. I usually have more action items than I can complete :(

rdpa:
@ndeloof I simply updated the dependency and tested locally and... everything seemed fine? I was able to add, update, and delete roles from every section and saw the changes reflected in what I was able to see job-wise. Is there any automated or more in-depth testing process I should look at?

oleg-nenashev:
There is not that much WebUI test automation for this feature. It is basically used for https://issues.jenkins-ci.org/browse/JENKINS-44472 only. You can test this feature manually. If it works well with the patch, let's go ahead with it. If not, we can consider reverting the change for now

oleg-nenashev:
I checked it manually, and it does not work

```
hudson-behavior.js:2952 Uncaught ReferenceError: Q is not defined
    at renderJobModal (manage-roles:202)
    at showJobsModal (manage-roles:186)
    at onSuccess (manage-roles:161)
    at klass.respondToReadyState (prototype.js:1657)
    at klass.onStateChange (prototype.js:1600)
    at XMLHttpRequest.<anonymous> (prototype.js:414)
```

oleg-nenashev:
FTR https://github.com/jenkinsci/role-strategy-plugin/pull/50 is the only code which uses the dependency

rdpa:
Planning on picking this back up soon

whut:
FYI, I found in https://github.com/jenkinsci/js-libs/tree/master/jquery-detached information that the approach used by jquery-detached in deprecated and instead this one should be used: https://github.com/jenkinsci/js-samples/blob/master/step-04-externalize-libs/HOW-IT-WORKS.md#configure-node-build-to-externalize-dependencies.

oleg-nenashev:
It is a "pick your poison" state, unfortunately. We have multiple ways to do it, there is no standard one ATM. Recent plugins usually use pure Maven Frontend Plugin + NPM 

Recent Examples: 

* https://github.com/jenkinsci/graphql-server-plugin by @halkeye 
* https://groups.google.com/forum/#!topic/jenkinsci-dev/Xt0cZhwTkS8 by @JackShen 
* https://github.com/jenkinsci/warnings-ng-plugin by @uhafner 

I started working on a proper fix for it a while ago, and I took approach from  https://github.com/jenkinsci/graphql-server-plugin . But we really need to define some best practices across the project

whut:
Dependency on old jquery-ui brings some additional dependencies on detached plugins, which in case of running Jenkins in a Docker container, triggers for example Ant and LDAP as dependency. Similar issue is reported with Green Balls in last comments in https://issues.jenkins-ci.org/browse/JENKINS-57528

whut:
Now I believe that the dependency on jQuery UI should be dropped, not "fixed", it just displays single dialog box, it can be easily replaced with simple "alert()"...

halkeye:
https://github.com/jenkinsci/role-strategy-plugin/blob/master/src/main/webapp/js/table.js#L58

So i'm not sure what immediateDescendants does, and can't find reference to the function, but everything else should be able to be converted to vanilla super easily.

There's some great examples at http://youmightnotneedjquery.com/

PR #99, opened 2019-07-29
Cleanup whitespaces in the code

karnauskas:
Just a cosmetic change, not altering plugin behaviour.


oleg-nenashev:
Thanks @karnauskas !

PR #110, opened 2019-11-07
[JENKINS-43058] - Replace obsolete slave/project terms in UI (part 2)

darxriggs:
This file has been missed in commit 765e489ab6ee21450c46e8083a4bb3661359eca9.

/CC @oleg-nenashev

### DIFF
--- a/src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/manage-roles.jelly
+++ b/src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/manage-roles.jelly
@@ -90,16 +90,14 @@
                 <f:block><st:include page="manage-global-roles.jelly" optional="true" /></f:block>
               </f:rowSet>
             </f:section>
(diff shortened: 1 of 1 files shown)


darxriggs:
@oleg-nenashev can you also merge it?

src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/manage-roles.jelly
```diff
@@ -90,16 +90,14 @@
...
               </f:rowSet>
             </f:section>
-            <f:section title="${%Project roles}">
+            <f:section title="${%Item roles}">
               <f:rowSet name="projectRoles">
                 <f:block><st:include page="manage-project-roles.jelly" optional="true" /></f:block>
               </f:rowSet>
             </f:section>
-            <f:section title="${%Slave roles}">
+            <f:section title="${%Node roles}">
```
oleg-nenashev: Agent would be better, because `master` is not supported in this field anyway. Fine with me

PR #114, opened 2019-12-10
Prevent errors when saving Role Strategy configurations after being configured with unmodifyable collections by JCasC

AndrewChubatiuk:
Fixes https://github.com/jenkinsci/configuration-as-code-plugin/issues/706


AndrewChubatiuk:
Moved fix to `retrieveRoleMap` method instead of `toMap`

AndrewChubatiuk:
@AbhyudayaSharma 
Updated `RoleMap` constructor to handle immutable sids

oleg-nenashev:
Tests passed, waiting for benchmarks which take a while due to whatever reason

PR #115, opened 2019-12-18
Cache Role Macros to improve the permission check performance

res0nance:
Signed-off-by: Raihaan Shouhell <raihaan.shouhell@autodesk.com>

Adds a cache for parsed macros. From what I gather this does static one time parsing so the cache doesn't seem like it needs expiry

jmh seems to indicate this as an overall perf win:

image


oleg-nenashev:
FYI @AbhyudayaSharma

PR #117, opened 2019-12-18
Add CODEOWNERS to automate review requests

oleg-nenashev:


AbhyudayaSharma:
Is it possible to skip the ci.jenkins.io builds on pull requests like this one which do not change the code?

oleg-nenashev:
We want to do it in order to reduce the workload on ci.jenkins.io. Here is one of the pull requests for it from @MarkEWaite : https://github.com/jenkins-infra/pipeline-library/pull/119

PR #118, opened 2020-01-02
Use size based-eviction for cache

res0nance:
Signed-off-by: Raihaan Shouhell <raihaan.shouhell@autodesk.com>

Based on this comment https://github.com/jenkinsci/jenkins/pull/4405#discussion_r362163614
looking into testing the performance of soft references vs size-based eviction methods.

PR #120, opened 2020-01-02
Fix `@since` fields in Javadoc + Use the new Role Strategy APIs with enum in the naming strategy and tests

res0nance:
Minor cleanups and update the since/todo comments

### DIFF
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
@@ -506,7 +506,7 @@ public class RoleMap {
     /**
      * Aborts the iterations.
      * The method can be used from RoleWalker callbacks to preemptively abort the execution loops on some conditions.
-     * @since TODO
+     * @since 2.14
      */
     public void abort() {
       this.shouldAbort=true;
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
@@ -506,7 +506,7 @@ public class RoleMap {
     /**
      * Aborts the iterations.
      * The method can be used from RoleWalker callbacks to preemptively abort the execution loops on some conditions.
-     * @since 2.14
+     * @since 2.10
      */
     public void abort() {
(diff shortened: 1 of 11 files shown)


src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java
```diff
@@ -506,7 +506,7 @@ protected Boolean hasPermission(Sid p, Permission permission) {
     /**
      * Aborts the iterations.
      * The method can be used from RoleWalker callbacks to preemptively abort the execution loops on some conditions.
-     * @since TODO
+     * @since 2.14
```
AbhyudayaSharma:
```suggestion
     * @since 2.10
```

Was introduced in https://github.com/jenkinsci/role-strategy-plugin/commit/f0e117d9b1ea56aec6fed40b4ced110c8b088818

src/main/java/org/jenkinsci/plugins/rolestrategy/RoleBasedProjectNamingStrategy.java
```diff
@@ -41,13 +43,13 @@ public void checkName(String name) throws Failure {
         if (auth instanceof RoleBasedAuthorizationStrategy){
             RoleBasedAuthorizationStrategy rbas = (RoleBasedAuthorizationStrategy) auth;
             //firstly check global role
-            SortedMap<Role, Set<String>> gRole = rbas.getGrantedRoles(RoleBasedAuthorizationStrategy.GLOBAL);
+            SortedMap<Role, Set<String>> gRole = rbas.getGrantedRoles(RoleType.Global);
```
oleg-nenashev: It may actually imply some performance differences, but it is fine in the case of this extension

res0nance: I double checked it, somehow performance is worse, and I'm doubting the bench tests for #119 and #118 they both show similar performance losses as what i saw from here. I took the tip of master and compared it to the benches for all 3 PRs. It seems like the baseline is off. Maybe an infra change is causing this Inaccuracy I might run the benches locally to verify.

In this exact case performance should improve or be of negligible difference since this reduces it by 2 calls `conversion from string to enum` and calling the non deprecated function.

PR #122, opened 2020-01-14
Upgrade the JCasC dependency to 1.35

timja:
JCasc had to create a new dependency that wasn't a `tests` classifier because maven doesn't allow you to bring dependencies through classifiers.

Upgrading here so that PCT works again,

Tracking:
https://github.com/jenkinsci/bom/pull/164

Note: We'll need a release for PCT please

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.46</version>
+        <version>3.55</version>
         <relativePath />
     </parent>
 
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Maybe it is too early to require the new JCasC version @timja

timja:
> Maybe it is too early to require the new JCasC version @timja

Why?

oleg-nenashev:
Potential breaking changes. When we update to 1.35, I would expect the plugin to also pick up new APIs and to export a proper schema. AFAIK it does not happen now

timja:
> Potential breaking changes. When we update to 1.35, I would expect the plugin to also pick up new APIs and to export a proper schema. AFAIK it does not happen now

The tests should cover for breaking changes,
> I would expect the plugin to also pick up new APIs and to export a proper schema

Is that a prerequisite for bumping the version? Once it's been upgraded someone could implement a schema for this plugin, note we haven't really looked into what a custom configurator needs to do (if anything)

oleg-nenashev:
As discussed at the JCasC meeting, I plan to proceed with this update.

PR #124, opened 2020-02-25
JENKINS-61205 System read support

timja:
Ref: https://github.com/jenkinsci/jenkins/pull/4524
<details>

<summary>Manage roles</summary>

image

</details>

<details>

<summary>Assign roles</summary>

image

</details>

The rest looks the same, are you fine with the extended-read-permission-plugin dep?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.55</version>
+        <version>4.0-beta-4</version>
         <relativePath />
     </parent>
 
@@ -42,9 +42,10 @@
     <properties>
         <revision>2.17</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.138.4</jenkins.version>
+        <jenkins.version>2.222-SNAPSHOT</jenkins.version>
         <java.level>8</java.level>
         <configuration-as-code.version>1.35</configuration-as-code.version>
+        <useBeta>true</useBeta>
     </properties>
 
     <repositories>
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>4.0-beta-4</version>
+        <version>3.55</version>
         <relativePath />
     </parent>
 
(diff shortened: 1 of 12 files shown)


pom.xml
```diff
@@ -75,6 +75,11 @@
             <artifactId>matrix-auth</artifactId>
             <version>2.2</version>
         </dependency>
+        <dependency>
+            <groupId>org.jvnet.hudson.plugins</groupId>
+            <artifactId>extended-read-permission</artifactId>
+            <version>3.1</version>
```
res0nance: Installing this will enable system-read and extended read for users who may not want those enabled. I don't see any harm in doing so but I'd like to highlight this

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig.java
```diff
@@ -69,6 +71,11 @@ public String getIconFileName() {
     return icon;
   }
 
+  @Override
+  public Permission getRequiredPermission() {
```
res0nance: Should add a comment to replace this with the regular system read and remove its dependency on the plugin once system read is in the baseline.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig.java
```diff
@@ -69,6 +71,11 @@ public String getIconFileName() {
...
+    return SystemReadPermission.SYSTEM_READ;
```
timja:
```suggestion
	// replace with Jenkins.SYSTEM_READ after baseline >= 2.222
    return SystemReadPermission.SYSTEM_READ;
```

jeffret-b:
A note for testing this to get around the Beta restriction for anyone who doesn't remember the incantation: Add `-Daccess-modifier-checker.skip=true` to get a testable build.

pom.xml
oleg-nenashev: We can invert the flag in the plugin if it helps @timja

timja:
> Not happening in short term (3 months?). I wonder whether we could move the required compatibility APIs to [jenkinsci/extended-read-permission-plugin](https://github.com/jenkinsci/extended-read-permission-plugin) or another plugin. IIUC we only need an equivalent of `l:isAdmin` there

what apis? isAdmin has existed since 2006

timja:
> Not happening in short term (3 months?). I wonder whether we could move the required compatibility APIs to [jenkinsci/extended-read-permission-plugin](https://github.com/jenkinsci/extended-read-permission-plugin) or another plugin. IIUC we only need an equivalent of `l:isAdmin` there

If you mean the pom version bump, I accidentally committed it, the only version bump is because of the extended-read baseline 2.138 LTS

timja:
@oleg-nenashev is a release coming soon?

oleg-nenashev:
yes

PR #130, opened 2020-05-14
Change rolewalk to use parallel streams

res0nance:
Try parallel streams with rolewalk. Will paste benchmarks when done


res0nance:
With parallel
image

res0nance:
With non parallel streams it also turns out to be significantly worse
image

PR #131, opened 2020-05-14
[SECURITY-410] Remove support for 'dangerous permissions'

timja:
equivalent of https://github.com/jenkinsci/matrix-auth-plugin/pull/80


oleg-nenashev:
I am merging it, because we have already had a major version bump

PR #136, opened 2020-07-25
Remove unneeded configuration in buildPlugin

timja:
Spotbugs reporting is being enabled by default in https://github.com/jenkins-infra/pipeline-library/pull/121, update to parent pom 4.x to use spotbugs instead of findbugs.


AbhyudayaSharma:
Looks like the benchmarks failed because of `java.lang.OutOfMemoryError: GC overhead limit exceeded`. Retriggering the build.

oleg-nenashev:
I have temporarily disabled benchmarks in #143 . We will still need to fix it later (soon)

timja:
:shipit: ?

oleg-nenashev (approved):

PR #139, opened 2020-08-13
Remove strict dependency on Extended Read Permission plugin

timja:
the core version is new enough that this is no longer needed

cc @oleg-nenashev @res0nance


oleg-nenashev (approved):

oleg-nenashev:
We need to update Jenkinsfile to make it work

res0nance (approved):

PR #140, opened 2020-08-21
set configuration category

jcoste-orange:
For Jenkins 2.226+, move the "Manage and Assign Roles" into the "Security" category.

Note that implementing this does not require bumping jenkins core dependency to a recent version, so the change proposed here is really minimal.


oleg-nenashev:
Test failure is unrelated.
Thanks @jcoste-orange !

oleg-nenashev (approved):
Better late than never, will go ahead and merge it

PR #142, opened 2020-09-30
[JENKINS-61470, JENKINS-61508] Invalidate cache for assign and unassign

blanked:
The cache (implemented in #115) is not invalidated when there is an assignment or un-assignment of roles to users, causing the users permissions to not be updated. This change adds invalidation for assignment and unassignment. I've also added web API tests.

### DIFF
--- a/src/test/java/com/michelin/cio/hudson/plugins/rolestrategy/APITest.java
+++ b/src/test/java/com/michelin/cio/hudson/plugins/rolestrategy/APITest.java
@@ -1,13 +1,10 @@
 package com.michelin.cio.hudson.plugins.rolestrategy;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.gargoylesoftware.htmlunit.HttpMethod;
 import com.gargoylesoftware.htmlunit.Page;
 import com.gargoylesoftware.htmlunit.WebRequest;
 import com.gargoylesoftware.htmlunit.util.NameValuePair;
 import com.synopsys.arc.jenkins.plugins.rolestrategy.RoleType;
-import hidden.jth.org.apache.http.HttpRequest;
-import hidden.jth.org.apache.http.auth.UsernamePasswordCredentials;
 import hudson.model.User;
 import org.acegisecurity.context.SecurityContext;
 import org.acegisecurity.context.SecurityContextHolder;
@@ -19,13 +16,9 @@ import org.jvnet.hudson.test.JenkinsRule;
 import org.jvnet.hudson.test.Issue;
 
 import java.io.IOException;
-import java.lang.reflect.Array;
 import java.net.HttpURLConnection;
-import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.SortedMap;
--- a/src/test/java/com/michelin/cio/hudson/plugins/rolestrategy/APITest.java
+++ b/src/test/java/com/michelin/cio/hudson/plugins/rolestrategy/APITest.java
@@ -1,6 +1,13 @@
 package com.michelin.cio.hudson.plugins.rolestrategy;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.gargoylesoftware.htmlunit.HttpMethod;
+import com.gargoylesoftware.htmlunit.Page;
+import com.gargoylesoftware.htmlunit.WebRequest;
+import com.gargoylesoftware.htmlunit.util.NameValuePair;
 import com.synopsys.arc.jenkins.plugins.rolestrategy.RoleType;
+import hidden.jth.org.apache.http.HttpRequest;
+import hidden.jth.org.apache.http.auth.UsernamePasswordCredentials;
 import hudson.model.User;
 import org.acegisecurity.context.SecurityContext;
 import org.acegisecurity.context.SecurityContextHolder;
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
Thanks @blanked ! There is quite a number of issues being addressed by this fix. I will review this PR and act on it ASAP

blanked:
Alright thanks!

src/test/java/com/michelin/cio/hudson/plugins/rolestrategy/APITest.java
```diff
@@ -0,0 +1,139 @@
...
+        securityContext.setAuthentication(User.getById("adminUser", true).impersonate());
+    }
+
+    @Test
+    public void testAddRole() throws IOException {
+        // Adding role via curl command
+        String roleName = "new-role";
+        String pattern = "test-folder.*";
+        String url = jenkinsRule.jenkins.getRootUrl() + "role-strategy/strategy/addRole";
+        String curlCmd = "curl -u adminUser:adminUser -X POST " + url + " --data type="+ RoleType.Project.getStringType()
```
res0nance, line 49: you probably want to do the test with something other than curl

res0nance (commented):
I think someone else was looking into this as well CC: @jvz

src/test/java/com/michelin/cio/hudson/plugins/rolestrategy/APITest.java
jvz, line 49: You can invoke the web methods directly here like `RoleBasedAuthorizationStrategy.getInstance().doAddRole()`, `doAssignRole()`, etc.

blanked: yeah i'll try to use the JenkinsRule WebClient instead

blanked: Oh I was thinking of simulating a web request. I'll use the `WebClient` from JenkinsRule instead? I'll make the change in a bit

src/test/java/com/michelin/cio/hudson/plugins/rolestrategy/APITest.java
```diff
@@ -0,0 +1,173 @@
...
+        }
+    }
+
+    /**
+     * Util method for executing a terminal command and getting the output
+     * @param cmd Command to be run
+     * @return String containing the terminal output
+     * @throws java.io.IOException
+     */
+    private static String execCmd(String cmd) throws java.io.IOException {
```
res0nance, line 169: Should be removed now

jvz (approved):
LGTM. Note the comment by res0nance.

oleg-nenashev (approved):
The code looks good to me. I suggest using https://issues.jenkins-ci.org/browse/JENKINS-61470 and https://issues.jenkins-ci.org/browse/JENKINS-61508 as a reference in the changelog. JENKINS-61508 is a better reference, because it describes the actual issue better.

oleg-nenashev:
I will go ahead and address changes so that it could be landed quickly

blanked:
> 
> 
> I will go ahead and address changes so that it could be landed quickly

Hey thanks man, forgot to clean it up and it was late. I made more clean up of the code. 
Btw, the CI job failed due to OOM on benchmarks, should we be concerned? Let me try running it locally.

oleg-nenashev:
Should be fixed in another PR

src/test/java/com/michelin/cio/hudson/plugins/rolestrategy/APITest.java
```diff
@@ -0,0 +1,159 @@
...
+        WebRequest request = new WebRequest(apiURL, HttpMethod.POST);
+        request.setRequestParameters(Arrays.asList(
+                new NameValuePair("type", RoleType.Project.getStringType()),
+                new NameValuePair("roleName", roleName),
+                new NameValuePair("sid", sid)
+        ));
+        Page page = webClient.getPage(request);
+        assertEquals("Testing if request is successful", HttpURLConnection.HTTP_OK, page.getWebResponse().getStatusCode());
+
+        // Verifying that alice is assigned to the role "new-role"
```
AbhyudayaSharma, line 117: I think the problem was that the cache was not being updated. The roles were being assigned but the cache was dirty and changes were not reflected. Are these tests also passing without invalidating the cache too?

AbhyudayaSharma (approved):
Hi @blanked. Thanks for fixing this. I tried to fix it several months but was not able to replicate the problem.

PS: If you would have created the PR today, it would have counted towards Hacktoberfest.

oleg-nenashev:
I started rebuild here: https://ci.jenkins.io/job/Plugins/job/role-strategy-plugin/job/PR-142/5/console

src/test/java/com/michelin/cio/hudson/plugins/rolestrategy/APITest.java
blanked: Oh actually I didn't run the tests before I made the change (sorry TDD). Looking at the test now, they might not be representative since I'm checking if the roles are updated via `getGrantedRoles`, which doesn't return the cache. I'm currently making a change to the test, will push in a bit.

blanked:
Ok I made an addition to the way the permissions are done: to use ACL (via `hasPermission`), which will ensure that the cache is used. I verified that the tests (after modification) fail on the old code (without invalidation) and passes with the changes.

oleg-nenashev (approved):
Thanks! Will go ahead and merge

PR #149, opened 2020-11-15
Performance: Use preauth filtering for matching names

res0nance:
Rewrite the getMatchingJobNames() implementation to use preauth filtering


oleg-nenashev (approved):
I wonder whether it shows some measurable performance improvement

PR #150, opened 2020-11-16
Swap from guava cache to caffeine

res0nance:
Guava provided from core is a cache library that probably can't be updated ever, use an updated library.

I re-ran the benchmarks as well

image

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -77,6 +77,11 @@
             <version>2.2</version>
         </dependency>
         <dependency>
+            <groupId>io.jenkins.plugins</groupId>
+            <artifactId>caffeine-api</artifactId>
+            <version>2.9.0-17.v79e9099a51e3</version>
+        </dependency>
+        <dependency>
             <groupId>org.hamcrest</groupId>
             <artifactId>hamcrest-junit</artifactId>
             <version>2.0.0.0</version>
@@ -99,11 +104,6 @@
             <artifactId>test-harness</artifactId>
             <version>${configuration-as-code.version}</version>
             <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>com.github.ben-manes.caffeine</groupId>
-            <artifactId>caffeine</artifactId>
-            <version>2.8.6</version>
         </dependency>
     </dependencies>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -100,5 +100,10 @@
             <version>${configuration-as-code.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.github.ben-manes.caffeine</groupId>
+            <artifactId>caffeine</artifactId>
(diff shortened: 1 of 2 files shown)


AbhyudayaSharma:
Wow! This is brilliant! 🎉

oleg-nenashev (changes requested):
Definitely +1 from me for switching, thanks @res0nance !

I doubt we can do it as is though. There will be a version collision with the Script Security plugin: https://github.com/jenkinsci/script-security-plugin/blob/f5c52156f0ea37a308c517f7a94784f96a1256ed/pom.xml#L74-L78 . Since the Script Security plugin is installed almost on every Jenkins instance, maybe it makes sense to add a dependency on the plugin or externalize the library as an API plugin.

CC @dwnusbaum

res0nance:
I'd vote for an API plugin, mostly due to if script-security drops caffeine for whatever reason this will just break role-strategy unintentionally. Perhaps Devin has a better idea.

dwnusbaum:
In general, I would prefer to not add new plugin dependencies to `script-security`, although an API plugin would probably be fine. We could try shading `caffeine` in `script-security` since it is not intended to be exposed to other plugins anyway, but if I recall correctly, that only works if we make `script-security` a multi-module project (unless we are shading the library directly into the plugin's sources), which I would also prefer to avoid, so I'm not sure what approach would be best here.

res0nance:
> In general, I would prefer to not add new plugin dependencies to `script-security`, although an API plugin would probably be fine. We could try shading `caffeine` in `script-security` since it is not intended to be exposed to other plugins anyway, but if I recall correctly, that only works if we make `script-security` a multi-module project (unless we are shading the library directly into the plugin's sources), which I would also prefer to avoid, so I'm not sure what approach would be best here.

We could add caffeine as a dep to core which means we would be able to move all of Core's guava cache usage to caffeine but updating caffeine in future might be tricky if has api breaks often.

ben-manes:
@res0nance The Caffeine APIs do not break often and honors semver. However, there will be a break in 3.0 which is in active development. Those changes are small quality-of-life improvements and will not impact most users.

dwnusbaum:
@res0nance I don't think we should ever intentionally add new dependencies to Jenkins core unless there is a way to hide them from plugins, since it just increases core's maintenance burden. To me an API plugin seems like the only realistic option here.

res0nance:
> @res0nance I don't think we should ever intentionally add new dependencies to Jenkins core unless there is a way to hide them from plugins, since it just increases core's maintenance burden. To me an API plugin seems like the only realistic option here.

Looks like James is adding a new api plugin https://github.com/jenkinsci/caffeine-api-plugin will adapt this PR once that is released

oleg-nenashev:
Thanks!

oleg-nenashev (approved):
Thanks @res0nance and @jtnord for making it possible!

PR #152, opened 2020-12-24
add code usage for groovy hook script/jenkins script console.

NagleZhang:
didn't find any code to do this, so write to readme to help more people.


oleg-nenashev (approved):
Thanks!

PR #165, opened 2021-12-21
Remove com.synopsys.arc.jenkins.plugins.rolestrategy.UserMacroExtension deprecated warnings

Conrad-T-Pino:
Remove com.synopsys.arc.jenkins.plugins.rolestrategy.UserMacroExtension deprecated warnings.

Remove internal API deprecated warnings in preparation for coming matrix-auth-3.0 API deprecated warnings.

- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

JIRA Issue: https://issues.jenkins.io/browse/JENKINS-67413
No relevant pull requests are known,
Changes limited to where a successful compile is a sufficient test.

### DIFF
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
@@ -26,7 +26,6 @@
 package com.michelin.cio.hudson.plugins.rolestrategy;
 
 import com.synopsys.arc.jenkins.plugins.rolestrategy.RoleType;
-import com.synopsys.arc.jenkins.plugins.rolestrategy.UserMacroExtension;
 import com.thoughtworks.xstream.converters.Converter;
 import com.thoughtworks.xstream.converters.MarshallingContext;
 import com.thoughtworks.xstream.converters.UnmarshallingContext;
@@ -640,12 +639,13 @@ public class RoleBasedAuthorizationStrategy extends AuthorizationStrategy {
      * Updates macro roles
      * @since 2.1.0
      */
+    @SuppressWarnings("deprecation")
     void renewMacroRoles()
     {
         //TODO: add mandatory roles
 
         // Check role extensions
-        for (UserMacroExtension userExt : UserMacroExtension.all())
+        for (final com.synopsys.arc.jenkins.plugins.rolestrategy.UserMacroExtension userExt : com.synopsys.arc.jenkins.plugins.rolestrategy.UserMacroExtension.all())
         {
             if (userExt.IsApplicable(RoleType.Global))
             {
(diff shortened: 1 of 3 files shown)


src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -640,12 +639,13 @@ public static RoleBasedAuthorizationStrategy getInstance() {
      * Updates macro roles
      * @since 2.1.0
      */
+    @SuppressWarnings("deprecation")
```
AbhyudayaSharma, line 642: Suppressing the deprecation here is fine. The other changes are not really necessary in my opinion.

AbhyudayaSharma (dismissed):
Hi. Thanks a lot for the pull request. However, I am unable to understand what this PR is trying to do. The only difference I see is suppressing a deprecation warning which on its own is fine. The other changes will not create any difference but can make the code harder to read. I would suggest limiting the changes to just the line highlighted in the review below.

Conrad-T-Pino:
`@SuppressWarnings("deprecation")` cannot be applied to import statements and `UserMacroExtension` class is deprecated.
Keeping import statements in place leaves three (3) instances of this warning in place:

1. `rolestrategy/RoleStrategyConfig.java:[30,53] com...UserMacroExtension in com.synopsys.arc.jenkins.plugins.rolestrategy has been deprecated`
2. `rolestrategy/RoleBasedAuthorizationStrategy.java:[29,53] com...UserMacroExtension in com.synopsys.arc.jenkins.plugins.rolestrategy has been deprecated`
3. `rolestrategy/macros/LoggedUserMacro.java:[28,53] com...UserMacroExtension in com.synopsys.arc.jenkins.plugins.rolestrategy has been deprecated`

* Unfortunately only deprecated class import warning workaround is the fully qualified class name; agreed it's ugly.
* IMO warnings should be calls to action and if no action is correct then suppression is the newbie-friendly course.

I'm training myself as new contributor and chose innocuous goal (zero warnings) as an exercise.  Your feedback and participation are valued.

AbhyudayaSharma:
Thanks for the reply. This is very interesting. I found https://bugs.openjdk.java.net/browse/JDK-8065613 which should ideally remove deprecation warnings for import statements. I checked this with JDK11 and JDK16 and I still get warnings; this should have been fixed with Java 9. Maybe this should be raised as a bug in javac given that the [JLS](https://docs.oracle.com/javase/specs/jls/se17/html/jls-9.html#jls-9.6.4.6) now states that:

> A Java compiler must produce a deprecation warning when an ordinarily deprecated program element is used (overridden, invoked, or referenced by name) in the declaration of a program element (whether explicitly or implicitly declared), **unless**:
> * ...
> * The use is within an import declaration that imports the ordinarily deprecated class, interface, or member; or ...

mawinter69:
UserMacros have been removed completely

PR #176, opened 2022-05-24
User/group regex pattern support

aerodinesh:
-      If I have to assign a role to all of my employees using one rule, using this change we can specify using regex

Sorry I am not well versed about open source policies. It would be very nice of anyone if they could take up this idea, and properly do the contribution.

Thanks again for all the community for selfless contributions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,16 +5,8 @@ updates:
   schedule:
     interval: weekly
   open-pull-requests-limit: 10
-  reviewers:
-  - oleg-nenashev
-  - runzexia
-  ignore:
-  - dependency-name: io.jenkins.plugins*
-    versions:
-    - ">= 0"
-  - dependency-name: org.jenkins-ci.main:jenkins-core
-    versions:
-    - ">= 0"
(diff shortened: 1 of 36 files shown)


AbhyudayaSharma (changes requested):
Hi. This PR will lead to a significant slowdown in the plugin's performance and will also change the existing behaviour. Could you not put all members into a group instead?

aerodinesh:
Oh, yeah I didnt consider performance impact, my bad. Yes, like you said we can create groups and use in regular authentication scenario, it works.
However in our setup we auth using google oauth2, and in that case somehow I could not find a way to create groups and pull users to them, maybe I need more info there. So, I have deployed this change locally and it helped me.
But yeah, only if it could help the community without side-effects we shud consider this :)
Thank you for the feedback.

aerodinesh:
Screenshot 2022-05-24 5 25 48 PM

This is how I used on our jenkins setup and it helped us to give access to our gsuite users through one rule.

aerodinesh:
ok, closing this.

PR #178, opened 2022-05-25
[JENKINS-67456] add check that pattern is valid

mawinter69:
performs a check of the pattern when the field looses focus
Additionally check when pressing the add button if the pattern is valid to avoid that an invalid pattern can be added at all.
see [JENKINS-67456](https://issues.jenkins.io/browse/JENKINS-67456)

### DIFF
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
@@ -74,6 +74,7 @@ import java.util.SortedMap;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
 
 import hudson.util.FormValidation;
 import jenkins.model.Jenkins;
@@ -938,6 +939,18 @@ public class RoleBasedAuthorizationStrategy extends AuthorizationStrategy {
             default:
                 return false;
         }
+    }
+
+    @RequirePOST
+    @Restricted(NoExternalUse.class)
+    public FormValidation doCheckPattern(@QueryParameter String value) {
+      try {
+          @SuppressWarnings("unused")
+          Pattern pattern = Pattern.compile(value);
+      } catch (PatternSyntaxException pse) {
+        return FormValidation.error(pse.getMessage());
+      }
+      return FormValidation.ok();
     }
 
     @RequirePOST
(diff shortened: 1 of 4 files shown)


src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -940,6 +941,18 @@ public boolean showPermission(String type, Permission p) {
         }
     }
 
+    @RequirePOST
+    @Restricted(NoExternalUse.class)
+    public FormValidation doCheckPattern(@QueryParameter String value) {
+      try {
+          @SuppressWarnings("unused")
+          Pattern pattern = Pattern.compile(value);
```
AbhyudayaSharma, line 949:
```suggestion
          Pattern.compile(value);
```

src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/manage-project-roles.jelly
```diff
@@ -148,6 +148,13 @@
             alert("Please enter a pattern");
             return;
           }
+          try {
+            new RegExp(pattern);
```
AbhyudayaSharma, line 152: This is a really interesting idea. But there are differences between `java.util.regex` and JavaScript `RegExp`. See https://en.wikipedia.org/wiki/Comparison_of_regular_expression_engines

mawinter69: ack

AbhyudayaSharma (approved):

NotMyFault (approved):

kuoruan:
The `escape` function is **deprecated** https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape
Can we use [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) instead?

NotMyFault:
> The `escape` function is **deprecated** https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape Can we use [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) instead?

Interesting, feel free to submit a new PR addressing the deprecation. I'm not a maintainer of this plugin.

PR #225, opened 2022-08-01
[JENKINS-19409] - Added option, which makes plugin to convert all SIDs to the lower-case (fixed merge conflicts)

ckreisl:
- [x] Link to relevant pull requests, esp. upstream and downstream changes

Hello, 
this based on PR #5 from 2018 and should resolve the merge conflicts. Just checked that everything is working as expected.
Right now I still get two checkstyle warnings:

``` 
[WARNING] src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java:[237,17] (naming) AbbreviationAsWordInName: Abbreviation in name 'getACL' must contain no more than '1' consecutive capital letters.
[WARNING] src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java:[556,14] (naming) AbbreviationAsWordInName: Abbreviation in name 'effectiveSID' must contain no more than '1' consecutive capital letters.
``` 

Feedback highly welcome since I'm not completely in Jenkins plugin development.

Cheers,
Chris

### DIFF
--- a/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
+++ b/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
@@ -138,7 +138,7 @@ public class RoleBasedAuthorizationStrategy extends AuthorizationStrategy {
   }
 
   public void setGlobalProperties(RoleStrategyProperties prop) {
-      globalProperties = prop;
+    globalProperties = prop;
   }
 
   /**
@@ -176,21 +176,6 @@ public class RoleBasedAuthorizationStrategy extends AuthorizationStrategy {
   }
 
   /**
-   private ACL getACL(String roleMapName, String itemName, RoleType roleType, AccessControlled item)
-   {
-     SidACL acl;
-     RoleMap roleMap = grantedRoles.get(roleMapName);
-     if(roleMap == null) {
-       acl = getRootACL();
-     }
-     else {
-       // Create a sub-RoleMap matching the project name, and create an inheriting from root ACL
-       acl = roleMap.newMatchingRoleMap(itemName).getACL(roleType, item, globalProperties.isConvertSidsToLowerCase()).newInheritingACL(getRootACL());
-     }
-     return acl;
-   }
-
-   /**
    * Get the specific ACL for projects.
    *
    * @param project The access-controlled project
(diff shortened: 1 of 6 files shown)


src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -895,6 +948,13 @@ public AuthorizationStrategy newInstance(StaplerRequest req, JSONObject formData
         strategy.assignRole(RoleType.Global, adminRole, getCurrentUser());
       }
 
+      // global properties
+      if (formData.containsKey("globalProperties")) {
```
mawinter69, line 952: use GLOBAL_PROPERTIES_NODE once moved

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -623,6 +659,9 @@ public void doGetMatchingAgents(@QueryParameter(required = true) String pattern,
    * </p>
    */
   public static class ConverterImpl implements Converter {
+
+    private static final String GLOBAL_PROPERTIES_NODE = "globalProperties";
```
mawinter69, line 663: Move that to the outer class so it can be used also in the descriptor

mawinter69 (changes requested):
This change completely ignores the fact that the security Realm now defines the IdStrategy (Which did not exist at the time of the first commit). This decides if comparison needs to be done case sensitive or case insensitive. Also the strategy can be different for users and groups so I think this needs to wait until JENKINS-68755 is implemented.
Ideally on a new Jenkins the plugin will use what the IdStrategy is using and only for old instances it will use the current case sensitive behavior, with the option to use the IdStrategy from the security realm.
Missing is also an integration with Configuration as Code probably. This needs to be covered by unit tests.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleBasedAuthorizationStrategy.java
```diff
@@ -128,6 +131,35 @@ public RoleBasedAuthorizationStrategy(Map<String, RoleMap> grantedRoles) {
...
+   */
+  public RoleBasedAuthorizationStrategy(Map<String, RoleMap> grantedRoles, RoleStrategyProperties prop) {
+    RoleMap map = grantedRoles.get(SLAVE);
+    agentRoles = map == null ? new RoleMap() : map;
+
+    map = grantedRoles.get(GLOBAL);
+    globalRoles = map == null ? new RoleMap() : map;
+
+    map = grantedRoles.get(PROJECT);
+    itemRoles = map == null ? new RoleMap() : map;
```
mawinter69, line 152: duplicated code with other constructor. The former should be deprecated and just forward to this one

ckreisl:
Thanks for the quick code review! Is there any specific timeline for integrating [JENKINS-68755](https://issues.jenkins.io/browse/JENKINS-68755)? I'll check on the review comments and on the mentioned [IdStrategy](https://javadoc.jenkins.io/jenkins/model/IdStrategy.html) class. Same applies for the JCasC implementation and unit tests.

mawinter69:
closed in favor of #332

PR #228, opened 2022-08-01
speed up user checking on assign roles

mawinter69:
The user checking is done sequentially by the FormChecker from core.
With many users this checking can take long time (reports are over 10
min for 300 users). This also depends on the securityRealm performance
and how many of these users really exist.
Confirmed with an AD and 300 users to take 10 to 12 min, with about 60%
of users not existing anymore.

To get a recognizable performance gain http2 needs to be enabled. When
this is done and running with 50 max parallel, the above scenario took
only around 1 min to load. Without http2, with 10 parallel connections
loading time went down to 5 min, with no further improvement with more
connections. It can even have bad impact in chrome, other browser
windows accesing the root of Jenkins was hanging with "waiting for
socket" in chrome.

### DIFF
--- a/src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/assign-roles.jelly
+++ b/src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/assign-roles.jelly
@@ -42,6 +42,9 @@
           <link rel="stylesheet" href="${rootURL}/plugin/role-strategy/css/role-strategy.css" type="text/css" />
           <script type="text/javascript" src="${rootURL}/plugin/role-strategy/js/table.js" />
           <script type="text/javascript" src="${rootURL}/plugin/role-strategy/js/assignFilters.js" />
+          <script>
+            FastFormChecker.parallelChecks = ${it.strategy.parallelChecks}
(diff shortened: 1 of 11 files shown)


src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/assign-roles.jelly
```diff
@@ -42,6 +42,9 @@
           <link rel="stylesheet" href="${rootURL}/plugin/role-strategy/css/role-strategy.css" type="text/css" />
           <script type="text/javascript" src="${rootURL}/plugin/role-strategy/js/table.js" />
           <script type="text/javascript" src="${rootURL}/plugin/role-strategy/js/assignFilters.js" />
+          <script>
+            FastFormChecker.parallelChecks = ${it.strategy.parallelChecks}
+          </script>
```
daniel-beck, line 47: Looks like this is the actual meat of this PR (once `FastFormChecker` is gone at least).

mawinter69:
This [PR-6951](https://github.com/jenkinsci/jenkins/pull/6951) from core will make this change here obsolete as it provides a solution that will work for other plugins as well

PR #230, opened 2022-08-05
[JENKINS-68755] Add support for explicit user and group assignment

mawinter69:
Adding support to explicitly assign a role to a user or a group. This avoids confusion when a name matches both. A warning will be shown on the role assignment page for ambiguous entries with the possibility to make this a user or a group assignment. 

❗ This is an incompatible change and after migration, going back to the previous version will require to undo the config changes. While the changes to the config.xml can be read by older versions, you can run into problems with duplicate entries.
Changes done on Configuration As Code yaml files need to be reverted to the old format before downgrading.

The new version can read existing configurations files (both from the config.xml and Casc). After startup of Jenkins with the new plugin version, the config.xml will be automatically saved with the new format.

The plugin APIs have significantly changed. Plugins that depend on role-strategy will probably need to be adjusted as well or might not work as expected (ownership-plugin, dynamic_extended_choice_parameter).

- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue


basil:
Required Configuration as Code plugin tests to be adapted: https://github.com/jenkinsci/configuration-as-code-plugin/pull/2357

PR #749, opened 2026-04-03
Re-design UI

timja:
TODO:
* [x] Get upstream dialogs change merged in that this PR requires
* [x] Check opening of the dialog is done right, I think its being done in JS when it should be done by setting the right jelly values
* [x] Check UI performance when lots of roles
* [x] Check groups work properly, e.g. display name and searching by display name
* [x] Adapt for system read
* [x] Test system read end 2 end once all other changes done
* [x] Check role manager permission works
* [x] Update permission templates to jelly dialog from js dialog
* [x] Switch to setting-subpage
* [x] Decide on navigation, e.g. permissions templates from roles page, should each action be duplicated or something else, Markus' feedback about subcategories could be useful here so we use the sidepanel instead of app-bar
* [x] The entry page name, heading is current role assignment but the title is something different
* [x] Should there be Save/ apply buttons or just save after every change like it currently does or should there be edit button that pops up a dialog
* [x] In the add dialog its just got plain radio and checkboxes, should that be enhanced?
* [x] Bottom button bar not working, fixed on role assignments page
* [x] Security realm update user and group name e.g ID to display name
  * Should work now, need to test groups
* [x] Review code in general

For later:
* Item based selection rather than just regex
* User / group autocompletion / validation on add - might look at it in this

https://github.com/user-attachments/assets/ffa2f2a9-5dfe-4fd1-9177-f44f1905cabd

### Testing done

Extensively manually tested including with different personas:
* Admin
* SystemRead
* Item roles admin
* Agent roles admin
* Item and Agent roles admin

Verified behaviour is correct across all of these.
I intend to create some playwright tests as well as there's 0 UI coverage at all in this repo (or in ATH), but that will likely be for another PR although I may stack it on this to get it going

### Submitter checklist
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed


timja:
There's probably some changes needed but its ready for a look from a design PoV.

* Code not ready for review yet
* Documentation probably needs updating
* Once we're happy with the design, data saving etc, I can look at ATH / PCT.

cc @mawinter69

timja:
Design crit review with Jan:

* [x] White text on light theme for template role patterns
* [x] edit dialog instead of selecting role assignment updating immediately - same for role and template page
GlOBAL ROLES to - Global roles
* [x] Side by side radio on assign role
* [x] Add templates button has disappeared
* [x] Can we move macros inline to the dialog somehow? or popup a dialog to the side?
* [x] Filter icon inconsistent between pages
* [ ] Check empty states, e.g. role page
* [ ] A testing framework for testing the UI of this, pipeline-graph-view and build-monitor are using PlayWright - might be worth adding this
* [ ] STRETCH: Selecting folders / jobs instead of regex
* [ ] STRETCH: Tippy dropdown? animation and shadows are wrong
  * Styling has been done but not animation, Jan found that really difficult, may need core improvements
* [ ] STRETCH: autocomplete on users and groups would be nice

General comment: Could simplify all the JS and make it more maintainable by bringing in e.g. React and/or TypeScript

timja:
@mawinter69 if you've got time to check the design it'd be appreciated I think its pretty much there.

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig.java
```diff
@@ -221,6 +310,276 @@
...
+  }
+
+  /**
+   * Called from the edit assignment dialog form submission.
+   * Replaces all role assignments for a given user/group.
+   */
+  @RequirePOST
+  @Restricted(NoExternalUse.class)
+  @SuppressWarnings("unchecked")
+  public void doEditAssignSubmit(StaplerRequest2 req, StaplerResponse2 rsp) throws IOException, ServletException {
```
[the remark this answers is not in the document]
timja: checked on the first line of parseAssignForm

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig.java
```diff
@@ -221,6 +223,125 @@
...
+    rsp.sendRedirect(req.getContextPath() + "/manage/role-strategy/");
+  }
+
+  /**
+   * Called from the add role dialog form submission.
+   */
+  @RequirePOST
+  @Restricted(NoExternalUse.class)
+  @SuppressWarnings("unchecked")
+  public void doAddRoleSubmit(StaplerRequest2 req, StaplerResponse2 rsp) throws IOException, ServletException {
```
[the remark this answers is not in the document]
timja: checked in parseRoleForm

src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig.java
```diff
@@ -221,6 +310,276 @@
...
+    rsp.sendRedirect(req.getContextPath() + "/manage/role-strategy/manage-roles");
+  }
+
+  /**
+   * Called from the edit role dialog form submission.
+   */
+  @RequirePOST
+  @Restricted(NoExternalUse.class)
+  @SuppressWarnings("unchecked")
+  public void doEditRoleSubmit(StaplerRequest2 req, StaplerResponse2 rsp) throws IOException, ServletException {
```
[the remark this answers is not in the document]
timja: checked in parseRoleForm

pom.xml
```diff
@@ -32,8 +32,8 @@
 
   <properties>
     <changelist>999999-SNAPSHOT</changelist>
-    <jenkins.baseline>2.479</jenkins.baseline>
-    <jenkins.version>${jenkins.baseline}.3</jenkins.version>
+    <jenkins.baseline>2.541</jenkins.baseline>
+    <jenkins.version>2.557-rc38182.209d0e525304</jenkins.version>
```
timja:
```suggestion
    <jenkins.version>2.560</jenkins.version>
```

pom.xml
```diff
@@ -32,8 +32,8 @@
 
   <properties>
     <changelist>999999-SNAPSHOT</changelist>
-    <jenkins.baseline>2.479</jenkins.baseline>
-    <jenkins.version>${jenkins.baseline}.3</jenkins.version>
+    <jenkins.baseline>2.541</jenkins.baseline>
```
timja:
```suggestion
    <jenkins.baseline>2.555</jenkins.baseline>
```

pom.xml
```diff
@@ -45,7 +45,7 @@
       <dependency>
         <groupId>io.jenkins.tools.bom</groupId>
         <artifactId>bom-${jenkins.baseline}.x</artifactId>
-        <version>5054.v620b_5d2b_d5e6</version>
+        <version>6210.v69ea_fd8a_f010</version>
```
timja:
```suggestion
        <version>6329.v403d8c87a_5ce</version>
```

seed-data.sh
```diff
@@ -0,0 +1,201 @@
+#!/bin/bash
```
timja, line 1: Useful script for performance testing

src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/add-role-dialog.jelly
```diff
@@ -0,0 +1,111 @@
...
+              </div>
+            </j:if>
+          </j:forEach>
+          <div class="rsp-assign-dialog__no-results jenkins-hidden" style="padding:1rem;text-align:center;color:var(--text-color-secondary);font-style:italic;">
+            ${%No matching permissions}
+          </div>
+        </div>
+      </f:entry>
+
+      <f:bottomButtonBar>
```
mawinter69:
```suggestion
      <f:bottomButtonBar borderless="true">
```

src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/add-template-dialog.jelly
```diff
@@ -0,0 +1,48 @@
...
+              </div>
+            </j:if>
+          </j:forEach>
+          <div class="rsp-assign-dialog__no-results jenkins-hidden" style="padding:1rem;text-align:center;color:var(--text-color-secondary);font-style:italic;">
+            ${%No matching permissions}
+          </div>
+        </div>
+      </f:entry>
+
+      <f:bottomButtonBar>
```
mawinter69:
```suggestion
      <f:bottomButtonBar borderless="true">
```

src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/assign-role-dialog.jelly
```diff
@@ -0,0 +1,84 @@
...
+          </j:if>
+
+          <div class="rsp-assign-dialog__no-results jenkins-hidden" style="padding:1rem;text-align:center;color:var(--text-color-secondary);font-style:italic;">
+            ${%No matching roles}
+          </div>
+        </div>
+
+      </f:entry>
+
+      <f:bottomButtonBar>
```
mawinter69:
```suggestion
      <f:bottomButtonBar borderless="true">
```

src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/edit-assign-dialog.jelly
```diff
@@ -0,0 +1,114 @@
...
+          </j:if>
+
+          <div class="rsp-assign-dialog__no-results jenkins-hidden" style="padding:1rem;text-align:center;color:var(--text-color-secondary);font-style:italic;">
+            ${%No matching roles}
+          </div>
+        </div>
+
+      </f:entry>
+
+      <f:bottomButtonBar>
```
mawinter69:
```suggestion
      <f:bottomButtonBar borderless="true">
```

src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/edit-role-dialog.jelly
```diff
@@ -0,0 +1,104 @@
...
+              </div>
+            </j:if>
+          </j:forEach>
+          <div class="rsp-assign-dialog__no-results jenkins-hidden" style="padding:1rem;text-align:center;color:var(--text-color-secondary);font-style:italic;">
+            ${%No matching permissions}
+          </div>
+        </div>
+      </f:entry>
+
+      <f:bottomButtonBar>
```
mawinter69:
```suggestion
      <f:bottomButtonBar borderless="true">
```

src/main/resources/com/michelin/cio/hudson/plugins/rolestrategy/RoleStrategyConfig/edit-template-dialog.jelly
```diff
@@ -0,0 +1,62 @@
...
+              </div>
+            </j:if>
+          </j:forEach>
+          <div class="rsp-assign-dialog__no-results jenkins-hidden" style="padding:1rem;text-align:center;color:var(--text-color-secondary);font-style:italic;">
+            ${%No matching permissions}
+          </div>
+        </div>
+      </f:entry>
+
+      <f:bottomButtonBar>
```
mawinter69:
```suggestion
      <f:bottomButtonBar borderless="true">
```

mawinter69 (changes requested):
Check boxes in the dialogs are touching each other.

Image

Maybe add some gap
```
.rsp-assign-dialog__group {
  display: flex;
  flex-direction: column;
  gap: 0.25rem;
}
```

mawinter69:
- When I assign a role and that user already has some roles assigned there is no error or warning shown, it seems to silently merge the roles. This is dangerous when when you have userids that are just like numbers, a simple typo could make you grant someone permissions that he should not have
- In the role management page the actually used userid/groupid is not visible also not as tooltip
- It would be nice when in the add role dialog the userid/groupis is directly resolved and the user/group displayname is shown, this helps to prevent giving the wrong user permissions
- In the dialogs to assign roles it is hard to know which permissions are connected to the roles. A tooltip with the permissions might help here, either hovering the role or via a small icon on the right
- Similarly when viewing the assigned roles of a user a tooltip might be helpful

mawinter69:
Seems in Matrix auth you have added small `?` icons showing the description of the permission as tooltip. Would be good to have that here as well

mawinter69:
In Matrix auth the permissions are directly editable. Maybe we can do that here as well (unless a template is used) and limit the edit dialog to item and agent roles to modify the pattern / template only.

mawinter69:
Handling for ambiguous assignments seems to be missing

mawinter69:
- the `lib/rolestrategy/dropdownList.jelly` taglib is not needed anymore 
- the braces.svg is not used anymore I think
- Adding a new role or assignment is currently a page reload, I think in matrix auth you made that purely in javascript

timja:
> Adding a new role or assignment is currently a page reload, I think in matrix auth you made that purely in javascript

Yes because saving is done when the security page as a whole is saved, whereas here its saved when the dialog closes.

It seemed clearer in this case to simplify everything by saving as part of the dialog, rather than saving every object on the page when you likely aren't changing much of it when adding e.g. new roles.

mawinter69:
> > Adding a new role or assignment is currently a page reload, I think in matrix auth you made that purely in javascript
> 
> Yes because saving is done when the security page as a whole is saved, whereas here its saved when the dialog closes.
> 
> It seemed clearer in this case to simplify everything by saving as part of the dialog, rather than saving every object on the page when you likely aren't changing much of it when adding e.g. new roles.

That would be an argument against my comment about being able to directly modify the permissions on the page. 
The APIs should be there that allow to add the role/assignment in javascript in the background. Would just need rendering the new entry once adding/editing was successful. When you want to add several users you and have that page reload after every user is not the best UX I think.

timja:
@mawinter69 would you be open to this being done in React.

I've got a port here: https://github.com/timja/role-strategy-plugin/pull/1
The state handling is far easier there, all the page edit, save etc.

Its got issues that I'm working through but just wanted to see if you'd be open to that?

mawinter69:
I have no experience with react, but if you think that this will make things easier then go ahead

timja:
> I have no experience with react, but if you think that this will make things easier then go ahead

Yeah its far easier to handle the interactivity and simpler code.

I'll extract some changes from this and start opening some PRs, react port is mostly there.

PR #765, opened 2026-07-22
Redesign manage roles page

timja:
TODO:
~* Performance with large sets of roles~ - I tried with a few 100 roles and it seems fine, can maybe paginate later if needed but it performed well for me.

https://github.com/user-attachments/assets/953250ea-f35c-4970-b936-181fe4a5c6ed

_AI Assisted, tested by me_

### Testing done

Unit, Playwright tests

Manually created global, agent and item roles

### Submitter checklist
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed


timja:
cc @janfaracik

janfaracik:
Pushed some minor changes to buttons + tabs, LGTM.

timja:
Thanks!

timja:
@mawinter69 could you take a look please?

src/main/frontend/common/components/Card.tsx
```diff
@@ -80,6 +80,10 @@ export function Card({
 
   const toggle = () => setExpanded((v) => !v);
   const onKeyDown = (e: KeyboardEvent<HTMLDivElement>) => {
+    // Only react to keys aimed at the header itself. Key events from focused
+    // buttons inside it (badges, actions) bubble up here, and preventDefault
+    // would swallow the native click those buttons synthesize on Enter/Space.
+    if (e.target !== e.currentTarget) return;
```
timja, line 86: accessibility issue I noticed that the view matching jobs couldn't be triggered via keyboard

timja:
Reminder @mawinter69

mawinter69:
- I think we should go to 2.568.3 with the Jenkins base
- For the tab bar, can we use the same classes that are used in the new dashboard UI? (i.e. `app-build-bar__tabs` `app-build-tabs, would require to use `a` instead of button. The buttons behave anyway a bit like links as they adjust the url)
- Getting from the Permission Template page back to the manage roles page is not directly possible. I think that is the logical next step after adding a new template. You either need to use the browser back or go the route via the assign page. So having a button there in the app-bar would be good I think (probably pre-existing since the redesign of the permission template page)

timja:
Done can you take another look please?

mawinter69:
One minor inconsistency
The buttons are ordered differently
image
image
Should be the same (do we have a convention what should be first?)

timja:
yes primary should be first

timja:
Done

mawinter69 (approved):

timja:
🚢 ?

